PR #16, opened 2026-07-17
Make inventory item viewport scrollable (fix Space Age overflow)

olafrose:
## Problem

The `InventoryDialog` is a fixed 404×442 panel. With Space Age, item groups (e.g. logistics) contain far more items than fit in the fixed height, and there was no clipping or scrolling — so the lower rows overflowed, were drawn over by the recipe panel, and could not be reached.

## Fix

In `packages/editor/src/UI/InventoryDialog.ts`:

- Add a PixiJS mask over the items viewport so content clips cleanly to the dialog bounds.
- Add mouse-wheel scrolling of the active group's items, clamped to the content bounds.
- Add a scrollbar thumb that reflects scroll position and hides when a group's contents fit.
- Refresh the scrollbar when switching groups (resets sensibly per group).

No changes to the dialog's outer dimensions, so surrounding layout is unaffected.

## Testing

- Verified in the running dev build (Space Age data):
  - Logistics (overflowing) → scrollbar appears, wheel scrolls, clamps at top and bottom, last row reachable.
  - Production (fits) → no scrollbar, all items visible.
  - Switching group tabs resets to top and refreshes the scrollbar.
- `npm run type-check` → 0 errors; `vp lint` / `vp fmt --check` clean for the changed file.

### DIFF
--- a/packages/editor/src/UI/InventoryDialog.ts
+++ b/packages/editor/src/UI/InventoryDialog.ts
@@ -1,4 +1,4 @@
-import { Container, Text } from 'pixi.js'
+import { Container, Graphics, Rectangle, Text } from 'pixi.js'
 import FD from '../core/factorioData'
 import G from '../common/globals'
 import F from './controls/functions'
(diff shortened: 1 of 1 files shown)


wormeyman:
Sorry for sitting on this for a week. The notification went past me in email and I only just caught it, which is on me, not on you.

Tanks for the fix. Logistics was the one group that actually overflowed (10 rows into an 8-row viewport, so the bottom two were unreachable), and this handles it cleanly. I ran the type-check gate, lint, and the unit tests against your branch and everything came back clean.

Merging now.

PR #45, opened 2026-07-26
fix: apply legacy name migrations only to the versions that need them

### LINKED ISSUE #40: Blueprint name migrations run unconditionally, rewriting live 2.0 entity names on load
## What happens

`nameMigrations` in `packages/editor/src/core/bpString.ts` is a table of legacy prototype renames. It is applied as a raw string replacement over the decoded blueprint JSON, before parsing:

```ts
const data = pako
    .inflate(decodedStr, { toText: true })
    .replace(nameMigrationsRegex, match => nameMigrations[match])
```

The table is grouped by the version each rename belongs to, but the conditions are **comments, not code**:

```ts
const nameMigrations: Record<string, string> = {
    // if (blueprintVersion < getFactorioVersion(0, 17, 0))
    '"raw-wood"': '"wood"',
    ...
    // if (blueprintVersion < getFactorioVersion(2, 0, 0))
    '"stack-inserter"': '"bulk-inserter"',
```

So every rename is applied to every blueprint regardless of the version it declares.

## Why it matters

Most of the 28 renames are harmless — the old name no longer exists in 2.0, so nothing legitimate can match. Two do not have that property. Their source names are **live prototypes in current data.json**, so a modern blueprint using them is silently rewritten:

| in the blueprint string | becomes | what the source name is today |
| --- | --- | --- |
| `stack-inserter` | `bulk-inserter` | live entity + item + recipe (reintroduced in 2.0) |
| `fusion-reactor-equipment` | `fission-reactor-equipment` | live item + recipe |

A Space Age blueprint containing stack inserters loads as bulk inserters, with no warning. Re-encoding then persists the substitution, so a copy/paste round trip through the editor changes the user's blueprint.

Because the replacement is textual over the whole JSON, it hits the name anywhere it appears — entity `name`, `recipe`, filters, icons, circuit signals.

## How it was found

Building a synthetic blueprint with one assembling machine per recipe for `tests/recipe-shapes.spec.ts` (#38). Two of the 653 recipes came back under a different name than they were placed with; those two are exactly the rows above. The spec asserts its recipe count loosely and documents why, so this is currently worked around rather than covered.

## Suggested fix

Restore the version conditions the comments describe, so each group only applies to blueprints declaring a version below its threshold. `decode()` would need the blueprint version before the rewrite, which means parsing first and migrating names on the parsed object rather than string-replacing the raw JSON — the textual approach cannot be made version-aware, since the version lives inside the string it is rewriting.

A narrower stopgap would be to drop just the two colliding rows, but that silently breaks genuinely pre-2.0 blueprints that contain them.

## Reproducing

```
echo '{"blueprint":{"item":"blueprint","version":562949957025792,"entities":[
  {"entity_number":1,"name":"stack-inserter","position":{"x":0.5,"y":0.5}}]}}'
```
Deflate + base64 with a leading `0`, load it, and inspect the entity: it is a `bulk-inserter`.

wormeyman:
Closes #40.

## The bug

`bpString.decode()` rewrote 28 legacy prototype names as a regex pass over the decoded JSON, before `JSON.parse`. The table was grouped by the version each rename belongs to, but the conditions were **comments rather than code**, so every rename applied to every blueprint. That could not be fixed in place: the version to test against lives inside the string being rewritten.

Most rows survived it because their source name no longer exists in 2.0, so nothing legitimate could match. Two do not have that property - `stack-inserter` and `fusion-reactor-equipment` are live prototypes again - so a current blueprint using either was silently rewritten on load, and re-encoding persisted the substitution.

## Measured impact

Every file in `wormeyman-tests/` declares 2.0.45 to 2.0.73 (1149 versioned blueprints and books), so **none of them should have been migrated at all**:

| | |
| --- | --- |
| `stack-inserter` occurrences in the corpus | **9535**, across 7 of the 11 files |
| what they loaded as | `bulk-inserter` - wrong name, wrong sprites, persisted on re-encode |
| other legacy source names present | none |

That corpus being 2.0-only is also why this went unnoticed: no test blueprint can reach a version-conditional branch.

## The fix

The migrations move onto the parsed object in `nameMigrations.ts`, grouped by threshold, scoped to the version each blueprint declares - so a 2.0 book holding a blueprint exported years earlier migrates only that entry, which the textual pass could not express. A blueprint with no version is treated as older than every threshold, matching what the textual pass did.

Note this is deliberately *not* the same call as `Blueprint.ts`'s version guards, which read `!== undefined` because `PaintBlueprintContainer` builds a version-less Blueprint for copy/paste that never comes through `decode()`.

Two deliberate divergences from the textual pass, both pinned by tests: a `recipe` field is dropped wherever it sits in its object rather than only when a field precedes it, and a legacy name inside an escaped quote in a label is no longer mangled. An entity's pre-2.0 `items` is a name -> count map, the one place the schema puts prototype names in object *keys*, so keys are renamed too.

## Tests, and why they are split

- **`packages/editor/src/core/nameMigrations.test.ts`** holds a frozen copy of the old table and asserts that a blueprint below *every* threshold still migrates identically. Only that case can be checked differentially - "all rows apply" was the right answer only there. Per-group behaviour is asserted directly.
- **`tests/name-migrations.spec.ts`** covers the same thing through the real decode path, where migrated names have to resolve against FD or the entity gets stripped. It builds its own blueprints, since the corpus cannot.

## Fixture changes, all reviewed

- **`sprite-data.json`** gains `stack-inserter` in the four halves that decode a blueprint. Purely additive - `bulk-inserter`'s digests are unchanged, because a renamed stack inserter drew an identical bulk inserter and deduped away. The `paintPreview` half **already held `stack-inserter` with these exact digests**, since placement from inventory never went through `decode()`; the fix makes the two paths agree, which is an independent witness that these are the right sprites.
- **`recipe-shapes.json`** gains `stack-inserter` and `fusion-reactor-equipment`, and the count assertion the bug forced to be loose is now exact.
- **`blueprint-round-trip`**'s `serializedHash` moved. Both position checksums and every count held, which is the signature of names changing under identical geometry.

## Verification

- `vp test` - 69 passed
- `npx playwright test` - 28 passed
- `npm run type-check:gate` - 81 errors, at baseline (unchanged)
- `vp lint` - no errors

Both commits are green standalone.

## Follow-ups found, not in scope

- Unknown *tile* names crash the load: `stripUnknownEntities` filters entities only, so a tile FD lacks reaches `TileContainer.generateSprite` and throws on `FD.tiles[name].variants`. Pre-existing and unrelated to migrations.
- `'straight-rail': 'legacy-straight-rail'` sits commented out precisely because the renames were ungated. It can be restored now, but it changes how existing pre-2.0 blueprints load, so it wants its own change.

https://claude.ai/code/session_016UMUmQi25ZDoHP7KbSK4Kn


wormeyman:
The two follow-ups from the PR description are now filed: #46 (unknown tile names crash the load) and #47 (restore the `straight-rail` row). Neither is a prerequisite for this change.

Issue #150, opened 2026-07-28
Paste places entities in blueprint order, so any placement rule needing other entities is order-dependent


wormeyman:
Split out of #141, which is closed. Nothing is broken today - this is a property of the paste loop that constrains what placement rules can ever be added, and it is on the board because it will bite the next one too.

## The property

`PaintBlueprintContainer.placeEntityContainer` iterates `this.entities` - a `Map`, so **insertion order**, which is the source blueprint's entity order - and each `PaintBlueprintEntityContainer.placeEntityContainer` asks `isAreaAvailable` against the grid **as it stands at that moment**.

So an entity's placement check can only see the entities that happened to come **before it in the blueprint's own entity list**. That order is whatever the game wrote when the blueprint was created; nothing normalises it.

## Why it matters

Any rule of the form "this entity may only go here if some *other* entity is already present" is order-dependent by construction. Measured for the one such rule that was actually costed (#141: an elevated rail needs a `rail-support` or `rail-ramp` supporting it), across 71 corpus blueprints holding 5922 elevated rails:

| radius tested | rails refused | blueprints affected |
| --- | --- | --- |
| 11 (the game's own `support_range`) | 1005 / 5922 | 71 of 71 |
| 13 | 580 | 64 |
| 18 | 226 | 57 |
| 25 | 62 | 36 |
| 1000 | **38** | **27** |

At radius 1000 the test degenerates to "has *any* support been placed yet" and it **still** refuses 38 rails in 27 of 71 blueprints. Real exports interleave supports after the rails they hold, so no radius rescues it - the variable is order, not distance.

Order-independently the same corpus is clean: **0 of 5922** elevated rails lack a support or ramp in their own blueprint.

## The shape of a fix

Two-pass placement in `placeEntityContainer`: place the entities that others depend on first - `rail-support` and `rail-ramp` are the known case - then everything else. That is a change to the paste loop, not to `isAreaAvailable`.

Worth being explicit that this has to land **before** any such rule, not with it. A dependency rule added on top of the current single pass makes pasting legal exports worse, which is the opposite of what a placement rule is for.

## Why this is not urgent

No rule in the editor currently depends on another entity being present, so nothing misbehaves today. `canCollide` (#136) is a filter over what is already in the area and does not care about order.

[...]

Issue #141, opened 2026-07-28
Elevated rails are accepted with no rail support beneath them


wormeyman:
Split out of #133 item 4, which is done. That item closed the *collision* half - the four `elevated-*` types are a collision layer rather than a geometry, and `canCollide` now models it (PR #136, fixture `tools/oracle/fixtures/elevated-rail-collision.json`). What it explicitly did not close is buildability, which is a different rule and is left here so it is on the board rather than only in a closed thread.

## The gap

The editor accepts an elevated rail with nothing holding it up. Factorio does not: `can_place_entity` refuses an elevated rail **everywhere**, on bare ground and over a chest alike, for exactly this reason.

This was measured as a side effect of #136 and is worth repeating, because it is the finding that voided a section of that probe:

> `create_entity` and `can_place_entity` disagree, and the gap is buildability rather than collision. `create_entity` built an elevated rail on bare ground with no support beneath it; `can_place_entity` refuses one everywhere. The over-a-chest sweep came back 0 accepted against a 0/8 empty-ground control and **voided its own section** - the entity it created was still standing, which is what let the collision direction be measured at all.

So the collision answer came from the prototype table (`collision_mask`), and buildability was never asked.

## Why this is not urgent

**No item places an elevated rail.** Nothing in `data.json` has one as its `place_result`, and the rail planner gives `straight-rail`, so the inventory cannot reach it and pipetting one hands back a ground rail. The only routes are pasting a selection that already contains one, or loading a blueprint that brought one - and a loaded blueprint never reaches `isAreaAvailable` at all. That leaves paste, where the elevated rails in the selection came from a real export and so already have their supports.

Which means the reachable failure is narrow: pasting a *partial* selection that clipped the supports out, or hand-building around one.

## What the work would be

Ask the game first, as `tools/oracle/README.md` says - and here the cheapest question is probably not a placement sweep. `rail-support` deliberately does **not** carry the elevated-rail collision layer, since a support has to overlap the rail it holds, so the relationship is not expressible as collision and something else in the prototype or the engine decides it. Worth checking `factorio-data` at the 2.0.73 tag before building a probe.

Then, if it is a real rule and worth enforcing, it belongs next to the other measured per-pair rules in `isAreaAvailable` rather than in the collision filter.

## Note on scope

[...]

PR #160, opened 2026-07-30
fix: make paste placement order-independent

### LINKED ISSUE #163: Paste drops legal entities when rectangular grid footprints overlap
## What happens

A pasted selection can contain entities that Factorio accepts together even though their rectangular footprints overlap in the editor's `PositionGrid`. `PaintBlueprintContainer.placeEntityContainer` used to check and place each entity in sequence, so the first pasted entity mutated the grid before the second was checked. The second entity was then silently dropped.

Two `curved-rail-a` entities two tiles apart are the measured case: Factorio accepts both, while the editor models each as a 2x6 rectangle and sees an overlap.

## Preview mismatch

`checkBuildable` computes the green/red tint for the whole paste against the destination grid at hover time, before any pasted entity is placed. The sequential placement loop could therefore render the second rail green and then refuse it after the first rail changed the grid.

## Scope

This is distinct from #150. The property needed here is that no entity within one paste blocks another: snapshot every placement decision against the destination grid, then execute those decisions. #150 needs the opposite visibility for dependency rules: selected entities such as rail supports must be placed first so later pasted entities can see them.

PR #160 implements the snapshot behaviour so the preview and result agree. Landing #150 later will require deliberately unwinding that plan-before-mutate invariant in favour of dependency-aware passes.

byalex33:
Fixes #163

## What changed

- Plan every pasted entity's create, replace, rotate, or blocked outcome before mutating the destination grid.
- Execute those recorded decisions in the existing transaction and wire-remapping flow.
- Record only the decision for replace/rotate and re-derive the live target during execution, so an earlier action cannot leave a later plan holding a destroyed `Entity`.
- Move the regression into `paste-placement.spec.ts` with a destination-collision negative control and real fast-replace/rotate coverage.
- Restore the `placeEntityContainer` return-contract documentation, name the planning invariant, require a placement argument, and update `CLAUDE.md`.

## Why snapshot instead of two-pass

`checkBuildable` computes the green/red tint for the whole paste against the destination grid at hover time, before any pasted entity is placed. The old sequential loop could therefore render the second entity green, place the first, then silently refuse the second against the newly mutated grid. Snapshotting the decisions makes the preview and the result agree.

That is intentionally different from #150. This change guarantees that no pasted entity blocks another; #150 needs selected dependencies such as rail supports to be placed first so later pasted entities can see them. #150 remains open, and implementing it will require deliberately unwinding this plan-before-mutate invariant in favour of dependency-aware passes.

## Root cause

`PaintBlueprintContainer` previously checked and placed each entity in one loop. Every successful placement changed `PositionGrid`, so later entities were judged against earlier entities from the same paste. The result depended on iteration order and could silently drop legal entities when the grid's rectangular footprints overlap.

The regression uses the measured rail case: two same-direction `curved-rail-a` entities two tiles apart are legal in Factorio, while their editor grid rectangles overlap. Before this change, pasting the pair added only one entity; now both are placed.

## Validation

- `vp test run`: 143 passed.
- Editor and website TypeScript projects: clean.
- Touched TypeScript files: lint clean; all touched files: format clean.
- Focused Playwright run covering `paste-placement`, `paste-wires`, `paste-modules`, `paste-entity-settings`, `paste-cross-type-settings`, and `elevated-rail-placement`: 34 passed.
- Full `playwright test` was run: 150 passed and 10 failed. Nine failures require the gitignored `wormeyman-tests/` corpus, which is not present in this checkout; the remaining numeric-input test passed 3/3 when its file was rerun. No paste-related test failed.


wormeyman (changes requested):
Reviewed at `ba8be86` against the local checkout.

The mechanical part of this is well done: the `'replace' | 'rotate' | 'create' | 'blocked'` union is a clean shape, the planning calls are genuinely read-only, and planning sits before `startTransaction` so the transaction still wraps exactly the mutations. Requesting changes on the issue linkage and four correctness/coverage items below.

One argument in the PR's favour that the body doesn't make, and it's the stronger one: `checkBuildable` computes the green/red tint from the grid **at hover time**, before anything is placed. So under the old sequential loop the tint and the outcome disagreed - the second rail rendered green and was then silently dropped. This change makes the preview and the result agree. That's a better justification than the entity count.

---

## 1. The approach contradicts what #150 asks for - please unlink it

#150's "shape of a fix" is explicit:

> Two-pass placement in `placeEntityContainer`: place the entities that others depend on first - `rail-support` and `rail-ramp` are the known case - then everything else.

The issue exists so a future rule of the form *"X may only go here if Y is already present"* becomes implementable. This PR does the opposite: by snapshotting every decision against the unmutated grid, no pasted entity can ever see any other pasted entity. Under this change the #141 rule that #150 was measuring would refuse **5922 of 5922** corpus elevated rails, not the 1005 the issue records at radius 11. The extension point #150 was filed to preserve is closed off, and implementing #150 later means unwinding this invariant first.

Both order-independence properties are legitimate - they're just different. This PR delivers "no pasted entity blocks another"; #150 asks for "a pasted entity can satisfy another's dependency".

Please:

- drop `Fixes #150` and open a separate issue for what this actually fixes ("paste drops legal entities when the grid's rectangular footprints overlap"), targeting that instead;
- say in the PR body why the snapshot approach was chosen over two-pass.

#150 stays open, with the note that #160's plan-before-mutate invariant has to be unwound before two-pass can land. The rail fix here is real and worth having - it just isn't #150.

## 2. Planned `replace`/`rotate` targets can go stale

The plan captures an `Entity` **object**, and execution can destroy it before a later plan uses it:

- Two paint entities at the same position in the same fast-replace group: the first `fastReplaceEntity` calls `removeEntity(target)` -> `onCreateOrRemoveEntity` -> `target.destroy()`. The second then runs `getEntityConnections`, `removeEntity` and `pasteSettings` against a destroyed object, and recreates at the stale `entity.position`.
- Replace-then-rotate on the same target: `placement.entity.direction = direction` on a destroyed entity.

Reaching this needs a source blueprint with stacked entities, so it's degenerate - but it's order-dependent behaviour introduced by an order-independence fix, and the plan/execute split makes it silent where the old single loop re-derived and found nothing.

Cheapest fix, and it removes the need to touch `Blueprint.ts` at all: **snapshot the decision, not the target.** Return a bare `{ type: 'replace' }` / `{ type: 'rotate' }` and re-derive the entity at execution time through the existing `checkFastReplaceableGroup` / `checkSameEntityAndDifferentDirection` calls. Re-deriving can only give the live entity or `undefined` (safely a no-op), and the ordering bug is about `isAreaAvailable`, which stays snapshotted.

That also disposes of the `fastReplaceEntity` signature change, which I'd rather not keep as written:

```ts
entity = this.entityPositionGrid.checkFastReplaceableGroup(name, direction, position)
```

A default computed from earlier parameters is opaque at the call site, and the passed entity is never checked against `name`/`direction`/`position`. Since the method then uses `entity.position` for the recreated entity, a mismatched argument silently builds in the wrong place. If the re-derive route isn't taken, prefer an explicitly named second entry point over an optional trailing arg.

## 3. The new test has no negative control

`expect(await entityCount(page)).toBe(4)` passes just as happily if `planPlacement` always returned `{ type: 'create' }`. Nothing in the suite currently pins that a pasted entity genuinely blocked by a **destination** entity is still refused, so the entire `isAreaAvailable` check could be deleted from the paste path and CI would stay green.

That's the mistake this codebase's conventions warn about directly - from CLAUDE.md on `elevated-rail-placement.spec.ts`: *"The refusals carry as much weight as the acceptances: a fix that ignored elevated rails entirely passes the first test and fails the next two."* Please add the sibling test.

Also newly load-bearing and uncovered: the `'replace'` and `'rotate'` branches. No test anywhere performs a real fast replace - `tests/position-grid.spec.ts:80` only asserts `checkFastReplaceableGroup` returns `undefined` for a self-match - so the new `fastReplaceEntity` parameter has zero coverage.

Two smaller test notes: the new case carries none of the surrounding file's explanatory comments (why `curved-rail-a`, why 2 tiles apart, why 4 entities, why 4 tiles right - `pixelsPerTile` being derived rather than hardcoded is correct and matches convention, it just isn't explained); and `editor-mode-input.spec.ts`'s header comment enumerates the modes it covers and isn't updated, so a paste-ordering regression may want its own file.

## 4. Deleted documentation, and no comment naming the new invariant

The 11-line JSDoc on `placeEntityContainer` - explaining that `undefined` is the normal outcome of three of the four paths, and why the wire-remapping caller skips it - was removed with nothing replacing it. `tests/paste-wires.spec.ts` is built on exactly that contract.

Worse for the future: nothing in the new code says *why* planning is separated from execution. A tidy-up that re-merges the two loops looks like a pure simplification and silently reinstates the bug. Please restore the old doc and add one on `planPlacement` naming the invariant - "planned against the grid as it stands before any of this paste is placed".

Related: `placeEntityContainer(placement = this.planPlacement())` creates a second path the only caller never takes. Making `placement` **required** encodes the invariant in the signature instead, which is strictly better than a comment.

## 5. CLAUDE.md not updated

Two entries go stale:

- Known Limitations: *"The refusals bite a user hand-building rails and a user **pasting** a selection, since `PaintBlueprintEntityContainer` asks about every entity it places"* - now true only for collisions with pre-existing entities, not within the paste.
- The `tests/editor-mode-input.spec.ts` entry doesn't mention the new test.

Every spec in this repo is described in CLAUDE.md; a PR adding one is expected to add its paragraph.

## 6. Please post a full Playwright run

CI here runs `vp check` and `vp test` (vitest) only - it never runs Playwright, so the browser suite is entirely on the author. The body reports `editor-mode-input.spec.ts` (13 tests), but paste behaviour changed and `paste-wires`, `paste-modules`, `paste-entity-settings` and `paste-cross-type-settings` all drive it. `elevated-rail-placement.spec.ts` calls `isAreaAvailable` through a hook rather than through paste so it should be unaffected, but that's worth confirming rather than assuming.

---

**Summary**

| | |
|---|---|
| Correctness | Sound for the common case; stale `replace`/`rotate` targets are a real if degenerate hole (2) |
| Design | Diverges from what #150 asked for without saying so (1) |
| Conventions | Deleted docs, no invariant comment, CLAUDE.md untouched (4, 5) |
| Tests | Positive case only; no refusal control; replace/rotate paths uncovered (3) |
| Performance | Negligible - one extra pass over the paste's entities |
| Security | No surface touched |

Taking (2)'s suggestion shrinks the diff and drops the `Blueprint.ts` change entirely, which I think leaves this in good shape.

byalex33:
Review updates are in `9c8d7c0e`.

Playwright results:

- Full `playwright test`: **150 passed, 10 failed**. Nine failures are the corpus-backed checks (`blueprint-round-trip`, `entity-accessors`, `entity-container-mappings`, `overlay-container`, `position-grid`, two `sprite-data` cases, `sprite-generation`, and `wire-connections`); this checkout does not have the gitignored `wormeyman-tests/` directory. The remaining failure was the numeric-input case in `text-input.spec.ts`, which passed **3/3** when that file was rerun.
- Focused paste/placement run: **34/34 passed** across `paste-placement`, `paste-wires`, `paste-modules`, `paste-entity-settings`, `paste-cross-type-settings`, and `elevated-rail-placement`.

Also green: `vp test run` (**143/143**), direct editor and website TypeScript checks, touched-file lint, touched-file format, and `git diff --check`.

wormeyman (approved):
Approving. All six points are addressed, and I re-verified each rather than reading the PR body - including building the degenerate stacked-entity case that point 2 was about, and mutation-checking the new spec four ways against a live dev server. Each of the four mutations kills exactly one test:

| Mutation | Test that fails |
|---|---|
| `planPlacement` returns `{type:'create'}` unconditionally | *a destination entity still blocks a paste* |
| `checkFastReplaceableGroup` arm deleted | *a planned fast replacement re-derives its live target* |
| `checkSameEntityAndDifferentDirection` arm deleted | *a planned rotation re-derives its live target* |
| base (unfixed) code | *a paste is planned before its own overlapping footprints change the grid* - `3` vs `4`, deterministic 3/3 |

So the negative control is real, the replace/rotate branches have genuine coverage, and the regression test genuinely regresses. Full suite independently confirmed at **170 passed, 0 failures** with the corpus present - which closes the 9 specs your own run reported as skipped, and shows the `text-input` case does not recur.

Point 2 took the better of the two routes: the decision is snapshotted and the target re-derived, so `Blueprint.ts` drops out of the diff entirely and the stale-object path is structurally unreachable rather than guarded. That is a smaller change than the one you started with, which is the right outcome.

Thanks for unlinking #150 and leaving the comment there recording that this PR's plan-before-mutate invariant has to be unwound before two-pass can land. That is the part that would have been lost.

Four follow-ups are in #181, none blocking. The one worth doing soon is the first: `checkBuildable` is now a second, hand-written copy of `planPlacement`'s three-question expression, and "the preview and the result agree" is this PR's own headline justification - so it is currently held true by two copies with nothing testing the equality. `this.planPlacement().type !== 'blocked'` makes it structural.

PR #197, opened 2026-08-10
Add a settings editor for display-panel entities

koenigstag:
Display-panel entity could be placed but had no editor at all - no way to set its icon or text, and the sprite-level icon rendering (EntitySprite/ spriteDataBuilder) only ever read whatever a blueprint import already carried.

Adds:
- An icon picker (DisplayPanelIcon, a Slot like Recipe/Modules) backed by a new Entity.acceptedDisplayPanelIcons - items, fluids, and virtual signals, matching what the game itself allows.
- A text field and an "Always show text in 'Alt-mode'" checkbox. Per the wiki, hovering always shows every line regardless of that setting; the checkbox only controls whether the first line stays visible above the entity without hovering. Rendered via two new OverlayContainer entry points (showEntityTooltip/hideEntityTooltip for the hover case, a block in createEntityInfo for the always-show case) and wired into EntityContainer's redraw/pointer handling, which previously had no display-panel awareness at all.
- A read-only list of a circuit-connected panel's control_behavior.parameters conditions - editing those live is a separate, larger feature and out of scope here.

InventoryDialog changes, needed by the icon picker rather than by display-panel itself:
- computeWidth widens the dialog when a filter populates more group tabs than the fixed 404px layout was designed for. The icon picker's items+fluids+signals filter is the first caller to hit this - up to 7 tabs against Space Age data, so the 5-groups-max layout used to hide the last tab(s) behind the dialog's own edge.
- The existing scroll viewport/scrollbar already handled tall groups, but the scrollbar thumb was pinned to a fixed VP_X + VP_W offset - correct at the old fixed width, but stranded well short of the actual right edge once computeWidth widens the dialog. Anchored it to `this.width` instead.
- A showRecipePanel flag (default on, matching every existing caller) lets the icon picker skip the recipe panel, which would otherwise render as a permanently empty bar - none of items/fluids/signals have a recipe.

### DIFF
--- a/packages/editor/src/UI/InventoryDialog.ts
+++ b/packages/editor/src/UI/InventoryDialog.ts
@@ -63,6 +63,9 @@ export class InventoryDialog extends Dialog {
     /** Scrollbar thumb rendered alongside the items viewport */
     private readonly m_ScrollThumb: Graphics
 
+    /** Whether the recipe panel was built - `setPosition` needs it after the constructor has returned */
+    private readonly m_ShowRecipePanel: boolean
+
     // Items viewport geometry (matches the layout comment above)
     private static readonly VP_X = 12
     private static readonly VP_Y = 126
@@ -70,15 +73,52 @@ export class InventoryDialog extends Dialog {
     private static readonly VP_H = 304
     private static readonly ROW_H = 38
 
+    /** Whether an item qualifies for a group tab, mirroring the population loop below */
+    private static itemQualifies(itemName: string, itemsFilter: string[] | undefined): boolean {
+        if (itemsFilter === undefined) {
+            const itemData = FD.items[itemName]
+            if (!itemData) return false
+            if (!itemData.place_result && !itemData.place_as_tile) return false
+            if (itemData.place_result && !FD.entities[itemData.place_result]) return false
+            return true
+        }
+        return itemsFilter.includes(itemName)
+    }
+
+    /**
+     * Widen the dialog if more group tabs need to fit than the 404px base
+     * layout was designed for. `DisplayPanelIcon` is the first caller whose
+     * filter (items + fluids + virtual signals) spans enough of
+     * `FD.inventoryLayout` to hit this - up to 7 tabs against Space Age data,
+     * where every other filtered caller stays within a handful of the base
+     * game's own groups.
+     */
+    private static computeWidth(itemsFilter: string[] | undefined): number {
+        let groupCount = 0
+        for (const group of FD.inventoryLayout) {
+            if (group.name === 'creative' && itemsFilter !== undefined) continue
+            const hasItems = group.subgroups.some(subgroup =>
+                subgroup.items.some(item => InventoryDialog.itemQualifies(item.name, itemsFilter))
+            )
+            if (hasItems) groupCount += 1
+        }
+        return Math.max(404, groupCount * 70 + 22)
+    }
+
     public constructor(
         title = 'Inventory',
         itemsFilter: string[] | undefined,
         // Not optional: picking an item is the whole point of the dialog, and all
         // five call sites pass one. Optional only made the invocation below a
         // type error, with nothing to do about it that was not a fiction.
-        selectedCallBack: (selectedItem: string) => void
+        selectedCallBack: (selectedItem: string) => void,
+        // Off for DisplayPanelIcon: it picks from items/fluids/signals with no
+        // recipe, so the panel would only ever show as a permanently empty bar.
+        showRecipePanel = true
     ) {
-        super(404, 442, title)
+        super(InventoryDialog.computeWidth(itemsFilter), 442, title)
+
+        this.m_ShowRecipePanel = showRecipePanel
 
         const D = InventoryDialog
 
@@ -227,27 +267,33 @@ export class InventoryDialog extends Dialog {
             }
         }
 
-        const recipePanel = new Container()
-        recipePanel.position.set(0, 442)
-        this.addChild(recipePanel)
-
-        const recipeBackground = F.DrawRectangle(
-            404,
-            78,
-            colors.dialog.background.color,
-            colors.dialog.background.alpha,
-            colors.dialog.background.border
-        )
-        recipeBackground.position.set(0, 0)
-        recipePanel.addChild(recipeBackground)
-
+        // Detached placeholders when the recipe panel is hidden - keeps
+        // updateRecipeVisualization a safe no-op rather than needing its own
+        // showRecipePanel check, since item buttons call it unconditionally.
         this.m_RecipeLabel = new Text({ text: '', style: styles.dialog.label })
-        this.m_RecipeLabel.position.set(12, 10)
-        recipePanel.addChild(this.m_RecipeLabel)
-
         this.m_RecipeContainer = new Container()
-        this.m_RecipeContainer.position.set(12, 36)
-        recipePanel.addChild(this.m_RecipeContainer)
+
+        if (this.m_ShowRecipePanel) {
+            const recipePanel = new Container()
+            recipePanel.position.set(0, 442)
+            this.addChild(recipePanel)
+
+            const recipeBackground = F.DrawRectangle(
+                this.width,
+                78,
+                colors.dialog.background.color,
+                colors.dialog.background.alpha,
+                colors.dialog.background.border
+            )
+            recipeBackground.position.set(0, 0)
+            recipePanel.addChild(recipeBackground)
+
+            this.m_RecipeLabel.position.set(12, 10)
+            recipePanel.addChild(this.m_RecipeLabel)
+
+            this.m_RecipeContainer.position.set(12, 36)
+            recipePanel.addChild(this.m_RecipeContainer)
+        }
 
         this.refreshScrollbar()
     }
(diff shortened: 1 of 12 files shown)


wormeyman:
CI was held pending maintainer approval, which is why this sat with no checks. It has been approved and the four checks pass.

I also ran the full Playwright suite against your branch merged into the current default branch. 175 of 176 pass, and I don't think the failure is a bug in your change.

`tests/overlay-container.spec.ts:78` now reports `display-panel: [-1, 1]`, where `EXPECTED_REAL` has no `display-panel` key at all. The reason it was missing is `summarise()` at lines 44-50, which drops any entity whose observations are all -1. Display panels drew no overlay before this PR, so the key never survived that filter. Now some of them draw one icon, so it does.

The fix is one line in `EXPECTED_REAL`, around line 145:

```ts
'display-panel': [-1, 1],
```

The `-1` stays in the list because some display panels still draw nothing.

Two things worth knowing beyond the one line:

CI here never runs Playwright. The green checks on this PR are `vp check` (format, lint, types) and the Rust exporter builds, so nothing in CI could have caught this. Reproducing it needs `npm run localpreview` in one terminal and `npx playwright test` in another.

CLAUDE.md treats that fixture as a fixed point rather than a refreshable snapshot. Adding the key is a deliberate "yes, display panels should draw an overlay now" rather than a regeneration, so please add the single line instead of rewriting the block from a test run.

Your branch is based on `0e06cbfa` and eight PRs landed on the default branch today, so a rebase would pick those up. It merges cleanly either way. `vp check` and `vp test` both pass on the merged tree, 218 files and 143 unit tests.

---

### If you are using Claude Code or another agent on this

Everything it needs, so you can hand it this section directly:

**The change.** In `tests/overlay-container.spec.ts`, add `'display-panel': [-1, 1],` to the `EXPECTED_REAL` object (starts at line 145). Keep the object's alphabetical order, so it goes between `'decider-combinator'` and `'electric-furnace'`. Single quotes, trailing comma, matching the entries around it. Do not change `summarise()` and do not touch any other key.

**Do not regenerate the fixture.** `EXPECTED_REAL` is a fixed point, not a snapshot to refresh. If a run disagrees with it, the run is the thing under suspicion. Only the one key above should change; if a diff touches others, stop and report that instead of accepting it.

**Verify.** Both dev servers must be up first, or every spec fails in a way that looks unrelated:

```bash
npm run localpreview                                    # repo root, starts both servers
npx playwright test tests/overlay-container.spec.ts     # in a second terminal
npx playwright test                                     # full suite, expect 176 passed
vp check && vp test                                     # format, lint, types, unit tests
```

`vp` must be on PATH. A bare `npm` inside the repo fails with `EBADDEVENGINES` unless `~/.vite-plus/bin` comes first, because `devEngines` pins npm `^12`.

**Worth reading before touching anything else here.** The repo's CLAUDE.md documents the conventions this code follows, and two of them bear on these files directly. `OverlayContainer.createEntityInfo` returns `Container | undefined`, where undefined is the normal case rather than a failure, since most entities draw nothing. And `Entity`'s accessors return `| undefined` on purpose where the blueprint format has the field optional, which `tests/entity-accessors.spec.ts` pins, so returning `[]` or `0` instead will fail that spec.

**Not required, but the obvious gap.** Nothing exercises the new editor itself. `tests/chest-editor.spec.ts` is the closest model for a dialog spec, including how it finds controls drawn inside a pixi dialog that has no DOM selector.

wormeyman:
One more thing before this goes in, and it is a request rather than a blocker on correctness: this needs a spec of its own.

The reason is specific to this repo rather than a general policy. Playwright never runs in CI here, so an editor can stop working entirely and every check stays green. That has happened: CLAUDE.md records the logistic chest editor sitting unreachable for two releases because its `factory.ts` case was commented out, which is the same line your change adds. Nothing noticed, because nothing was looking.

Your diff is also doing more than registering a dialog. `EntityContainer` now hides `entityInfo` when a display panel has `displayPanelText`, and that visibility swap is exactly the kind of thing the existing suite cannot see. `overlay-container.spec.ts` tallies what `createEntityInfo` returns, not whether the container it produced is visible, so the branch you added would go on passing if it stopped working.

`tests/chest-editor.spec.ts` is the model to copy. It is the closest thing here to what you need, because it is the only spec that clicks a control drawn inside a pixi dialog, which has no DOM selector to target.

Happy to look at a draft if any of it fights you.

---

### If you are using Claude Code or another agent on this

**The task.** Write `tests/display-panel-editor.spec.ts` covering the editor added in this PR. Model it on `tests/chest-editor.spec.ts`, which solves the same problems.

**How a dialog spec works here.** There is no DOM to query, so positions come from hooks and layout constants:

- `openEditorOn` in `chest-editor.spec.ts:155` is the routine to copy. It reads `window.__fbe_test.entityScreenPosition(n)`, moves the pointer to `at.x, at.y + 240` first and then to `at.y`, then clicks. The two-step move is deliberate: the editor recomputes hover on pointer movement, so moving straight to coordinates the pointer already occupies changes nothing and the click lands in `NONE` mode.
- `window.__fbe_test.topDialogBounds()` gives the dialog's position. Control coordinates are `dialog.x + <constant>`, where the constants mirror the editor's own layout, the way `FILTERS_X` and `SLOT_PITCH` at `chest-editor.spec.ts:52-54` mirror `ChestEditor` and `Filters`. Derive yours from `DisplayPanelEditor` the same way, and say in a comment which source they mirror.
- `window.__fbe_test.openDialogCount()` and `window.__fbe_test.editorMode()` assert the dialog opened and the mode is `EDIT`.

**Two things that will waste a run if missed.**

Call `suppressOverlays(page)` from `tests/helpers/overlays.ts` before `page.goto`, not after. Two DOM overlays sit above the canvas and swallow clicks: toasts bottom right and the dat.gui pane bottom left. A lost click looks exactly like a broken setter.

Build the blueprint with `tests/helpers/encode-blueprint.ts` and choose the entity numbers yourself. Do not use a corpus file. Loading re-centres the view, so assert positions as offsets rather than the coordinates you encoded, and keep entities close together or the far one is off screen and hovers nothing.

**What to cover.** The dialog opens on a display panel and closes again. A text edit reaches the entity. The `EntityContainer` branch this PR adds, where `entityInfo` is hidden once `displayPanelText` is set, since no existing spec can see that. Assert against the serialized blueprint rather than the model where a setter might write a correct value into the wrong shape.

**Prove the spec can fail.** Break the thing it covers, run it, confirm it goes red, then put it back. A spec that passes against broken code is worse than none, and this suite has several tests whose comments exist only to record which mutation they were checked against.

**Run it.** `npm run localpreview` in one terminal, `npx playwright test tests/display-panel-editor.spec.ts` in another, then the full suite. `vp check && vp test` before pushing.

wormeyman:
Reviewed the new spec and ran the full Playwright suite on the merged tree. This is what I was asking for, and the `EXPECTED_REAL` line in `545f0e0e` is exactly right too: one line, correctly placed, nothing else touched.

## The spec reaches the dialog

That was the whole point of asking. Nothing in CI here runs Playwright, so an editor can stop working with every check green, which is how the logistic chest editor sat unreachable for two releases behind the same `factory.ts` line this PR adds.

I mutation-checked it three ways and each one kills exactly the right tests:

| Mutation | Result |
| --- | --- |
| remove `case 'display-panel':` from `factory.ts` | tests 1 and 2 fail, test 3 passes |
| remove `entityInfo.visible = false` in `pointerOverEventHandler` | only test 3 fails (`Expected: false, Received: true`) |
| remove `entityInfo.visible = true` in `pointerOutEventHandler` | only test 3 fails (`Expected: true, Received: false`) |

The `entityInfoVisible` hook is the right way to get at that. `overlay-container.spec.ts` tallies what `createEntityInfo` returns, so the visibility swap was invisible to the whole suite before this. Asserting test 2 against the serialized blueprint rather than the model is also correct, for the reason your comment gives.

## One test fails, and you cannot see it

`tests/display-panel-editor.spec.ts:158` presses `Control+A`. On macOS that is the emacs "beginning of line" binding rather than select-all, so the typed text lands in front of the old text and the assertion gets `"New textOld text"`. `ControlOrMeta+A` fixes it, verified over three repeats. It is the only `Control+A` in the suite, so nothing else was masking it.

No need for another round on your side. I will apply the one-liner when I merge.

Full suite on the merged tree: 178 passed, 1 failed (that one), 5.4 minutes. 179/179 with the fix. `vp check` and `vp test` are both clean, and the branch merges onto the current tip without conflicts.

## What is still uncovered

Writing these is on me rather than a third round for you:

* the alt-mode checkbox, which is the only control drawn in both editor layouts and which nothing clicks. `always_show` appears in the spec only as fixture input.
* the icon picker, along with the `showRecipePanel` parameter you threaded through `createInventory` into `InventoryDialog`.
* the connected branch. `DisplayPanelEditor` returns early when `generateConnector` is true, so `createConditionDisplay` and `displayPanelParameters` never run. The fixture has no wires.

Two things I could not settle by measuring, so I am flagging them rather than claiming them:

* `pointerOutEventHandler` re-reads `displayPanelText` before restoring. If the text is cleared between hover-in and hover-out, the restore is skipped and the tooltip stays up. Nothing exposes the tooltip container, so I could not prove that path is reachable. A handler that undoes its pair unconditionally would be safer either way.
* that early return means a wired panel has no text or icon field at all. Whether that matches the game is a factorio-data question rather than something to reason out, so it goes on the follow-up list.

Merging. Thanks for the two rounds on this.

Issue #193, opened 2026-08-05
Blueprint with a wire between a splitter and a belt fails to load (upstream teoxoy#273) - very likely fixed by #191, needs verifying


wormeyman:
Mirrors upstream <https://github.com/teoxoy/factorio-blueprint-editor/issues/273>, reported by @gth on 2026-07-10 and unanswered there. Filing here because this fork is affected too.

## Their report

> When pasting a blueprint, the parser fails if a splitter is wired to a belt (e.g. in order to control splitter settings based on belt contents).

Repro: place a splitter and a belt, run a red or green wire between them, export, paste into FBE.

## What their repro string actually contains

Decoded offline (no editor needed), their attached blueprint is version 2.0.77 and holds:

- entity 2: `express-splitter`
- entities 1, 3-8: `express-transport-belt`
- `wires: [[2, 2, 4, 2]]` - a **green circuit wire from the splitter to a belt**

## Why this is almost certainly already fixed

This is the same failure mode as the bug found while replacing the test corpus (#186), fixed in PR #191:

`getCircuitConnector` in `packages/editor/src/core/factorioData.ts` never named `splitter`, so a splitter fell to `default:` and answered undefined. Nothing catches around `WiresContainer.add` - the call is a bare `forEach` inside `BlueprintContainer.initBP`, and the only `try` in `Editor.loadBlueprint` sits above it around `getChildIndex` - so a single wired splitter threw `Could not find the wire connection point!` and the **whole blueprint** failed to load rather than one sprite. That matches their symptom exactly: a parser/paste failure rather than a missing splitter.

`getMaxWireDistance` had the same omission, with the quieter cost of a reach of 0 drawing wires at alpha 0.3.

Both are fixed on `public-blueprint-corpus` (PR #191), and `tests/splitter-wires.spec.ts` pins both halves.

## What still needs doing

**Verify, do not assume.** Nobody has loaded *this* blueprint string. Two specific reasons it might not be wholly covered:

1. Their wire uses connector **2** (green). The regression test and the corpus blueprint that exposed the bug both use connector **1** (red). The connector index is not what was missing - the whole `case` was - so red and green should behave identically, but that has not been checked against this string.
2. Their string is version 2.0.77 and holds an `express-splitter`; the corpus covers `turbo-splitter` and `express-splitter` wired, so this should be in range.

[...]

Issue #186, opened 2026-08-05
sprite-data.spec.ts fails outright without the gitignored wormeyman-tests corpus, so 57% of its digests are unrunnable on any other clone


wormeyman:
## The shape

`tests/__fixtures__/sprite-data.json` is the characterization harness for `packages/editor/src/core/spriteDataBuilder.ts` - the largest file in the repo (2,738 lines) and the most-churned (47 commits in 12 months). It holds **3,645 digests** across five halves, and `tests/sprite-data.spec.ts` is its only consumer (repo-wide grep for `sprite-data.json` returns that one file).

Two of the five tests source their blueprints from the corpus:

- `tests/sprite-data.spec.ts:124` - "real blueprints generate the same sprite data with no position grid" (`noGridReal`, 668 digests)
- `tests/sprite-data.spec.ts:195` - "real blueprints generate the sprite data they generated before" (`real`, 1,427 digests)

Both open with:

```ts
const files = discoverBlueprintFiles()
expect(files.length).toBeGreaterThan(0)
```

`discoverBlueprintFiles()` reads `wormeyman-tests/`, which is gitignored at `.gitignore:10`:

```
/wormeyman-tests
```

## Why it matters

On any clone that lacks the corpus, those two tests do not degrade to reduced coverage - they **fail at the guard**, before generating anything. That is 2,095 of 3,645 digests, 57% of the harness, unavailable to any contributor.

And it is specifically the half that carries the interesting coverage. The spec says so itself at `tests/sprite-data.spec.ts:188-193`:

> The synthetic blueprint spaces entities out so none of them touch, so it never reaches the branches that read neighbours - pipe junctions, belt corners, underground pairs, heat pipe and wall connection sprites, gate rail bases. Those are a large part of the file, which is why the real bases carry the other half of the coverage here.

So without the corpus there is no junction coverage at all. Concretely, `real.pipe` holds 17 digests and `real["heat-pipe"]` holds 39, and those are the only ones reaching the 15-branch connection ladders in `draw_pipe` (`spriteDataBuilder.ts:2110`) and `draw_heat_pipe` (`:1648`). The three corpus-free halves reach only the single fallback branch for those entities.

## This is not the same as "CI doesn't run Playwright"

[...]

PR #205, opened 2026-08-10
test: assert both wire switches cover every type in data.json

### LINKED ISSUE #188: getCircuitConnector and getMaxWireDistance have no guard that they cover every type in data.json
Both switches in `packages/editor/src/core/factorioData.ts` are keyed by entity type and have to be extended by hand when a prototype gains a circuit connector. Nothing checks that they are complete.

That gap cost a real bug, found while replacing the test corpus (#186): `splitter` was missing from both. The two halves failed very differently.

- `getCircuitConnector` fell to `default:` and answered undefined. Nothing catches around `WiresContainer.add` - the call is in a bare `forEach` inside `BlueprintContainer.initBP`, and the only `try` in `Editor.loadBlueprint` sits above it around `getChildIndex` - so a single wired splitter threw `Could not find the wire connection point!` and **the whole blueprint failed to load**. The throw lands before `addChildAt`, so the container is never attached: everything else on screen went with it.
- `getMaxWireDistance` fell to `default: return 0`, and a reach of 0 only makes `WiresContainer.createWire` draw the wire at alpha 0.3, as though the ends were too far apart. Wrong and completely silent.

**Both switches are complete today** - measured against `packages/exporter/data/output/data.json`, zero types carrying `circuit_connector` are unhandled by the first, and zero types carrying `circuit_wire_max_distance` are unhandled by the second. So this is not a live defect; it is the absence of anything that would catch the next one.

`tests/splitter-wires.spec.ts` pins the splitter case specifically, but per-type specs are the wrong shape for this. The durable fix is one data-driven unit test that reads `data.json`, collects every type declaring either field, and asserts each is named by the corresponding switch. It needs no browser and no FD, so it belongs in vitest rather than Playwright, and it would have caught the splitter case before anyone loaded a blueprint.

Worth noting what would NOT have caught it: the corpus. 578 blueprints' worth of fixtures stayed green because no file in the old corpus contained a wired splitter.

wormeyman:
> **Was stacked on #204, which has merged** (f371796) and brought the `unit` vitest project this test runs in. Rebased onto `wormeyman-space-age-support`, so this is now the single new file it said it would collapse to, and CI runs on it.

`getCircuitConnector` and `getMaxWireDistance` are per-type switches extended by hand when a prototype gains a circuit connector, and nothing checked they were complete. That gap cost a real bug (#186): `splitter` was missing from both, and the two halves failed nothing alike.

- `getCircuitConnector` fell to `default:` and answered undefined. Nothing catches around `WiresContainer.add` - a bare `forEach` inside `BlueprintContainer.initBP`, with the only `try` in `Editor.loadBlueprint` sitting above it around `getChildIndex` - so one wired splitter threw `Could not find the wire connection point!` and took **the whole blueprint** down. The throw lands before `addChildAt`, so the container is never attached and everything on screen goes with it.
- `getMaxWireDistance` fell to `default: return 0`, and a reach of 0 only makes `WiresContainer.createWire` draw the wire at alpha 0.3, as though the ends were out of range. Wrong, and completely silent.

Both are complete today, so this is not a fix. It is what catches the next one before a user does, which #193 is the worked example of: that bug reached a user upstream and sat unanswered for a month.

## How it asks

**Behaviourally, not by parsing the source for `case` labels.** It calls the real function with the real prototype and checks the answer is not the default one. Reading the text would answer "is this type named", which is a proxy; this answers "does a wire have somewhere to go". It also survives a reformat and catches an arm that names a type and still hands back nothing.

Measured against `data.json` rather than assumed:

- 80 entities declare a circuit connector - 78 at the top level, and the two rail signals under `ground_picture_set` instead. Those same two also carry `elevated_picture_set`, which the editor does not read and this does not assert.
- 84 declare `circuit_wire_max_distance`, none below 9, so the 0 a missing case answers is unambiguous. Asserted for **equality** rather than non-zero, which also catches an arm added to the wrong group.
- `electric-pole` and `power-switch` declare neither, keeping their reach in `maximum_wire_distance` / `wire_max_distance`, so they get their own test - the data-driven half is blind to them.

**A vitest test under `tests/` rather than beside the code**, and both halves are deliberate. Neither function reads `FD`, so no browser is needed and it runs in CI where Playwright does not. But `packages/editor` narrows its tsconfig `types` back to typed-factorio on purpose, so browser code cannot reach node globals, and this needs `fs`; the root config is the only one here carrying node types.

## Verification

`vp check .` clean, `vp test` 154 passing across 12 files. Playwright unchanged at 176 tests in 40 files, and does not collect this file. No production code changed.

Mutation-checked five ways, each killing exactly one test:

| Mutation | Test that fails |
| --- | --- |
| Delete `case 'splitter'` from `getCircuitConnector` | the connector test, naming all four splitters |
| Delete `case 'splitter'` from `getMaxWireDistance` | the distance test, `got 0, declares 9` |
| `getCircuitConnector` default returns a connector | the pipe control |
| `getMaxWireDistance` default returns 9 | the pipe control |
| Delete the `electric-pole` arm | the test for the two types whose reach lives elsewhere |

The first two are the actual #186 bug, reproduced and caught:

```
AssertionError: these fall to the default arm and so lose their whole blueprint
+   "splitter (type splitter)",
+   "fast-splitter (type splitter)",
+   "express-splitter (type splitter)",
+   "turbo-splitter (type splitter)",
```

Closes #188

### DIFF
--- a/tests/wire-switch-completeness.test.ts
+++ b/tests/wire-switch-completeness.test.ts
@@ -1,0 +1,213 @@
+import { describe, expect, it } from 'vite-plus/test'
+import * as fs from 'fs'
+import * as path from 'path'
+import { EntityWithOwnerPrototype } from 'factorio:prototype'
+import { getCircuitConnector, getMaxWireDistance } from '../packages/editor/src/core/factorioData'
+
+/*
+    `getCircuitConnector` and `getMaxWireDistance` are per-type switches that a
+    new prototype has to be added to by hand, and nothing checked they were
+    complete. That gap cost a real bug (#186): `splitter` was missing from both,
+    and the two halves failed nothing alike.
+
+    - `getCircuitConnector` fell to `default:` and answered undefined. Nothing
+      catches around `WiresContainer.add` - it is a bare `forEach` inside
+      `BlueprintContainer.initBP`, and the only `try` in `Editor.loadBlueprint`
+      sits above it around `getChildIndex` - so one wired splitter threw
+      "Could not find the wire connection point!" and took the *whole blueprint*
+      down. The throw lands before `addChildAt`, so the container is never
+      attached and everything else on screen goes with it.
+    - `getMaxWireDistance` fell to `default: return 0`, and a reach of 0 only
+      makes `WiresContainer.createWire` draw the wire at alpha 0.3, as though the
+      ends were out of range. Wrong, and completely silent.
+
+    Both are complete today, so this is not a live defect - it is the thing that
+    catches the next one before a user does. What would *not* have caught the
+    splitter is the corpus: 578 blueprints stayed green because no file in it
+    held a wired splitter, and `tests/splitter-wires.spec.ts` pins that one type
+    by name, which is the wrong shape for a question about coverage.
+
+    Three things about where this sits and how it asks the question.
+
+    It is a vitest test living in tests/, which is a directory of Playwright
+    specs, and both halves of that are deliberate. Neither function under test
+    reads `FD` - both take a prototype directly - so nothing has to be loaded and
+    no browser is needed, which also means it can run in CI, where Playwright
+    does not. But it cannot live next to the code it guards: `packages/editor`
+    narrows its tsconfig `types` back to typed-factorio alone, on purpose, so
+    that browser code cannot reach node globals, and this needs `fs`. The root
+    config is the only one here carrying node types, so a test that reads a file
+    off disk sits under a directory that config owns whatever it is testing.
+
+    It reads the real `packages/exporter/data/output/data.json`, the same file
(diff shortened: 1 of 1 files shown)


wormeyman:
CI does not run on this PR: `.github/workflows/ci.yml` triggers only on pull requests whose base is `wormeyman-space-age-support`, and this one is based on `corpus-stray-file-guard` (#204). GitHub will retarget it automatically when #204 merges, and CI fires then.

Ran CI's three commands locally on this branch in the meantime:

```
vp check .              218 files formatted, 0 warnings or errors
vp test                 154 passed across 12 files
npm run type-check:gate 0 errors (at baseline 0)
```

Plus `npx playwright test --list`: 176 tests in 40 files, unchanged, and this file is not collected by Playwright.

Issue #119, opened 2026-07-27
Paste-settings specs intermittently paste nothing under a full run


wormeyman:
Split out of #115, where it was measured but not fixed.

## What happens

Under a full `npx playwright test`, a paste-settings test occasionally fails with the pasted value simply absent:

```
Error: expect(received).toEqual(expected)
Expected: {"a": 0.5, "b": 0.1, "g": 0.1, "r": 0.9}
Received: undefined
  at tests/paste-cross-type-settings.spec.ts:311
```

That is a paste that **did nothing** rather than a setter that wrote the wrong thing. Every affected test passes on its own.

## It is not caused by any one change

Measured across six full runs on 2026-07-27, three either side of #115's change:

| | runs | runs with a failure | which |
| --- | --- | --- | --- |
| with the #115 change | 3 | 2 | `paste-cross-type: the colour goes the other way too` ×2, `paste-cross-type: the editor offers the paste only for the pairs it can act on` |
| baseline, that change stashed | 3 | 1 | `paste-entity-settings: a cargo wagon carries its inventory limit and slot filters together` |

Different tests each time, same family, same shape. Roughly one run in three.

## What has already been ruled out

- **Not the hover.** Both specs' `hoverEntity` already does the step-off-then-on move and then `waitForFunction` on `hoveredEntityNumber`, so the editor has confirmed the hover before either click is dispatched. `paste-entity-settings.spec.ts` documents adding that wait for exactly this symptom, and it did fix the common case - the cargo wagon test used to fail about one run in two.
- **Not the #111 startup race.** That was `window.__fbe_test` appearing before `loadData` finished, which produced `Cannot read properties of undefined` from wherever the data was first touched. This is a clean load followed by a gesture that lands on nothing.
- **Not #109's HMR theory.** No file is edited during a run, and that mechanism turned out not to reproduce at all (see #117).

## Where to look

The race is *after* the confirmed hover, in `pasteSettings`:

```ts
await hoverEntity(page, from)
await page.keyboard.down('Shift')
await page.mouse.down({ button: 'right' })   // copy source captured here
await page.mouse.up({ button: 'right' })
await hoverEntity(page, to)
await page.mouse.down()                       // paste applied here
await page.mouse.up()
await page.keyboard.up('Shift')
```

Two candidates, neither confirmed:

[...]

PR #221, opened 2026-08-12
Add a ToolsPanel with show details/import string/export string/export image/undo/redo UI actions

koenigstag:
## Summary
- Expose import-replace, import-append, export-to-string and export-to-image — actions that only ever existed as keyboard shortcuts (paste, Ctrl+Shift+V, copy, Ctrl+S) — as buttons, plus Undo/Redo (Ctrl+Z/Ctrl+Y).
- Added ImportDialog (textarea + Replace/Append buttons with descriptions) and ExportDialog (pre-filled, pre-selected textarea) as a clipboard-free alternative, for browsers that block `navigator.clipboard` or a string that needs to be seen/edited.
- Renamed WiresPanel to ToolsPanel, since it now holds import/export/undo/redo alongside the wire slots.

image

image

## Layout notes
- The panel wraps into a column-major grid instead of a single wide row — a wide single row runs part of it under `.toasts-container` (`position: fixed; right: 0`), confirmed by clicking a widened-panel slot and finding the click reached the toast instead of the button.
- Undo/Redo call `G.bp.history.undo()/redo()` directly (same as their keybinds in Editor.ts); the four clipboard/file actions go through a new `QuickActions` interface injected via `Editor.init()`, mirroring the existing `Logger` injection.
- Icons: game GUI sprites (`import_slot`/`export_slot`) and the `signal-anticlockwise-circle-arrow`/`signal-clockwise-circle-arrow` virtual signals for Undo/Redo — the only undo/redo-shaped icons anywhere in vanilla Factorio's data.

## Test plan
- [x] `vp check` / `vp test` pass
- [ ] Manually verified all quick actions, both dialogs, and Undo/Redo end-to-end against the dev server


wormeyman:
Thanks for this, and the panel-wrapping investigation in the description is the kind of thing that usually goes unrecorded.

One thing I'd like to settle before it lands. `Editor.init` becomes:

```ts
init(canvas, quickActions, logger?)
```

`quickActions` is required and it sits ahead of the optional `logger`. That's the exported entry point of `@fbe/editor`, so an existing `init(canvas, createToast)` call now binds a logger to the `quickActions` parameter. From TypeScript that's a compile error, which is the good case. From JS it's silent.

There's exactly one caller in this repo, so nothing is broken today. What gives me pause is that it's a published entry point, and the ordering makes the break sharper than it needs to be. Either of these would avoid it: put `quickActions` after `logger` and make both optional, or take a single options object.

I don't have a strong preference between them, and if you think the break is fine given the single caller, say so and I'll merge as is.

One other thing, not blocking: the description doesn't mention `BookDialog`, which is around 267 lines and a feature in its own right. It made the PR harder to review than it needed to be.

The branch is current with the base now and CI is green on it, including the browser suite, which started running in CI this week.

koenigstag:
Fixed the mentioned - split BookDialog into a separate PR per your scope note, and switched Editor.init to a single options object per your API-signature note.

wormeyman:
I measured this in a browser before reading much of it, because the panel got wider and this repo has been bitten by that before (#119).

At 1280x720, with the toasts that a blueprint load raises:

```
panel          x 861..1073
toast column   x 960..1280
```

`document.elementFromPoint` at each slot's centre:

| slot | what gets the click |
| --- | --- |
| ALT, copper-wire, import, green-wire | `CANVAS` |
| export, red-wire, undo, redo, export-image | `DIV.toasts-container` |

Five of the nine buttons do nothing while a toast is up. Toasts fire on every blueprint load, on 5s and 10s timers, and the welcome toast lives 30 seconds. That is the state right after someone loads a blueprint, which is when they are most likely to reach for undo.

The comment on `ROWS` says wrapping into two rows keeps the panel clear of `.toasts-container`. I measured the single-row case too and it is worse, so that half holds. The wrap does not close the gap though: 212px starting at x 861 still reaches 113px into the 320px column.

I also measured base, so the report is fair about what this PR introduces. `green-wire` is already covered there, 1 slot of 3. So the bug predates this PR and this makes it about five times bigger.

The fix I would take is not in this PR at all. `.toasts-container` in `packages/website/src/index.css` has no `pointer-events` rule, so the container eats clicks across its whole 320px column whether or not a toast sits under the pointer:

```css
.toasts-container { pointer-events: none; }
.toasts-toast { pointer-events: auto; }
```

That clears all five, clears the pre-existing `green-wire` case, keeps toasts clickable to dismiss, and shrinks what `tests/helpers/overlays.ts` has to work around. It is a base-branch change, so it belongs in its own PR rather than here.

Three smaller things.

`exportString` and `exportImage` return `boolean`, but `QuickActions` types them `() => void` and `ToolsPanel` drops the result. Clicking export-image on an empty blueprint does nothing and says nothing. The keybind used that boolean to mean "not handled".

`G.app.ticker.add()` in `generateSlots()` is never removed, and `generateSlots` is public, so a second call stacks a second callback. `WiresPanel` had the same public method with nothing to leak, so this is new.

`functions.ts` now has two separate `import ... from 'factorio:prototype'` lines.

Nothing under `tests/` touches any of this. `QuickActions` is a new seam with two callers per action, which is the shape that breaks quietly: the keybind path and the button path can drift apart with every check green.

The panel itself is a good idea and the icon research behind it is the right kind of work. It is the geometry I want fixed before it lands.

wormeyman:
Correcting myself on the fix I recommended above. The count was right and the remedy was wrong.

My probe read `elementFromPoint` 500ms after the load. Toasts slide in from the right, so it caught them mid-animation, still short of their resting x. That is why four slots came back as `DIV.toasts-container`: the container was exposed because the toast had not arrived over it yet.

Measured again, after settling on toast identity and geometry together:

```
container   x 960..1280, y 588..720
toast-2     x 960..1280, y 596..661
toast-1     x 960..1280, y 669..715
```

A toast is `position: relative` inside a 320px container, so a settled toast spans the **full** column. What is left uncovered is three margin bands totalling about 21px of height.

The slots, once settled:

| slot | hit |
| --- | --- |
| export | `DIV.toasts-toast toasts-warning` |
| red-wire | `DIV.toasts-toast toasts-success` |
| undo | `DIV.toasts-toast toasts-warning` |
| redo | `SPAN.toasts-text` |
| export-image | `DIV.toasts-toast toasts-warning` |

So it is the toast rectangles that eat the clicks, not the space around them, and `pointer-events: none` on `.toasts-container` alone would have bought you the 21px of margin and the 300ms of slide-in. Not a fix.

Still five of nine, and still `green-wire` on the current default branch, so nothing about the problem changed.

The general lesson, which is the part worth keeping: **an overlay measurement taken before the stack settles is measuring the animation.** The same point reports the container or the toast depending on when you look.

## Where it goes from here

Filed as #228 and opened a PR against base. `pointer-events` inherits, so one declaration on the container reaches the toasts too, which makes ordinary toasts click-through. The two toasts created with `timeout: Infinity` keep their pointer events, since clicking is the only way to dismiss those and both callers throw straight afterwards anyway.

**What that does for this PR:** the five buttons stop losing their clicks. What it does not do is stop a toast being drawn over them. They are still covered, just no longer dead. If the panel should not sit under the toast column at all, that is a layout question rather than a pointer one, and it is worth deciding on purpose rather than inheriting from where `WiresPanel` happened to sit.

Everything else in my earlier comment stands.

wormeyman:
Re-reviewed at `410946eb`. The three small items are fixed: `QuickActions` types the two export actions as returning `boolean`, the ticker callback is removed before a new one is added, and the duplicate `factorio:prototype` import is gone. `Editor.init` taking an options object reads well, and the doc on `EditorInitOptions` makes the argument for it better than my comment did.

I drove this one rather than only reading it, since last time the interesting part was geometry. Numbers below are measured at `410946eb` with both dev servers up.

## Merge base in. It closes the toast problem this panel's layout was built around

None of your three open PRs contain `b39292f9` (#229). This branch is four base commits behind.

Measured on the branch as it stands, 1280x720, one live toast, `document.elementFromPoint` at each slot centre:

| slot | hit |
| --- | --- |
| ALT, copper, import, green | `CANVAS` |
| export, red, undo, redo, export-image | `DIV.toasts-toast` / `SPAN.toasts-text` |

Five of nine, the same five I measured on 13 August. Then I dropped in base's `index.css` and `toasts.ts` and changed nothing else:

| slot | hit |
| --- | --- |
| all nine | `CANVAS` |

So merging base fixes every one of them, and no change is needed here for it.

Two consequences worth noting. The comment on `ROWS` says wrapping into two rows "keeps the panel as narrow as the action count allows" so it stays clear of `.toasts-container`. After the merge that justification is stale: the clicks get through either way. The wrap may still be worth keeping for looks, but it is no longer load-bearing, and leaving the old reason in place will send the next person hunting for a constraint that is gone. The second is that clicks getting through is not the same as the panel being visible. Five slots still sit under a toast for the first 30 seconds after a load. Whether that matters is a layout call, and it is worth making on purpose rather than inheriting from where `WiresPanel` happened to sit.

## The Import dialog cannot import the largest blueprint in the repo

`TextInput`'s `maxLength` setter writes the real `maxlength` attribute, and `ImportDialog` passes `2 ** 20`. Browsers apply `maxlength` to a paste, not only to typing.

Pasting `test-blueprints/EARN/pocket-base-space-age-v22.1.2.txt`:

```
source string length : 2417093
textarea maxlength   : 1048576
textarea value length: 1048576
characters lost      : 1368517
decoding the result  : throws
```

The user gets a "corrupt blueprint string" toast that says nothing about truncation. The same string pastes fine with Ctrl+V on the canvas, so the dialog is worse than the shortcut it was added to expose.

`ExportDialog` sets the same limit at line 28 and gets away with it, because assigning `.value` from a script ignores `maxlength`. Measured: 410,133 characters present in the export field with the limit still set. So the fix is only needed on the import side, but I would drop the cap on both rather than leave one that works by accident.

## The panel runs off screen below about 866 px of viewport width

`setPosition` offsets from a hardcoded 442 quickbar width with no clamp. The panel went from 136 px wide to 212, which moves the threshold:

```
1280x720  panel 861..1073   fits
1000x800  panel 721..933    fits
 900x800  panel 671..883    fits
 800x800  panel 621..833    33px off screen, export-image slot centre at 803 in an 800px viewport
 700x800  panel 571..783    83px off screen
```

`isMobile.any` is about the device, not the width, so a desktop browser at 800 px wide draws the panel and loses the last slot with no way to reach it. `WiresPanel` only clipped below 714. It is a narrow band and I am not treating it as a blocker, but a clamp costs one line.

## Smaller things

**`generateSlots()` still is not safely re-callable.** The comment at lines 165-170 says it is, and that is why the ticker is removed first. But the method never clears `slotsContainer`. `QuickbarPanel.generateSlots` is re-callable only because its caller, `changeActiveQuickbar`, calls `removeChildren()` before it. A second call here leaves 18 slots stacked in pairs, each with a live handler. So the ticker guard defends a case the rest of the method still breaks. Either clear the container or drop the claim from the comment.

**The ticker callback is never removed on destroy.** `Panel.destroy()` exists precisely to unhook the resize listener. `ToolsPanel` adds a second global subscription and does not override `destroy`. Nothing reaches it today, since `Editor.init` runs once, but the commented-out `changeQuickbarRows` in `UIContainer` is the shape that would.

**`CELL_COUNT = WIRES.length + 6` is `cells.length` written down twice.** It sizes the background before `cells` exists, so only a person can keep them equal. Add a tenth action without bumping the 6 and it renders half outside the panel, with no type error, no lint error, and nothing under `tests/` able to see it. Build `cells` first and size the panel from `cells.length`.

**`QuickActions.exportString` is dead.** Declared, documented, implemented in the website and injected, and nothing in `packages/` reads it. The Export slot calls `G.UI.toggleExportDialog()`, and the website's `copy` listener calls its own local function. Either wire it up or drop it, since as it stands every future host of `Editor.init` has to implement it for nothing.

**The wire order changed.** `cells` uses `WIRES[0]`, `WIRES[2]`, `WIRES[1]`, so the panel now reads copper, green, red where `WiresPanel` read copper, red, green. The layout comment does describe the pairing it produces, so I think this is deliberate, but it does not say it is a change. Confirm it is what you meant, since the middle slot moving is the sort of thing muscle memory notices.

**`F.CreateUtilitySpriteIcon` and `F.CreateIcon` are called from the constructor with no catch above them.** `ToolsPanel` is built inside `new UIContainer()` inside `Editor.init`, so one missing sprite or signal name takes down the whole editor rather than one blank slot. That is the rule CLAUDE.md states for `need()`: find the nearest try/catch above the read before assuming a throw is affordable. This one is not hypothetical either, in the sense that #222 has a live crash of exactly this shape from an unguarded `CreateIcon`.

## What I checked and found clean

I expected Escape to stop working while the Export dialog is open, since `select()` focuses the textarea and `Editor.ts`'s keydown handler returns early for `HTMLTextAreaElement`. It does not happen. Measured over 2.5 seconds after opening Export, `document.activeElement` stays `CANVAS` throughout while the textarea is `display: block` with 410,133 characters in it, and Escape closes the dialog. The `focus()` call does not take, most likely because the element is still hidden at the moment it runs. Worth knowing so nobody "fixes" it later and reintroduces the trap.

`tests/quick-actions.spec.ts` is a real addition and covers the seam I was worried about. One note on it: the Append test reads `editorMode()` straight after `page.mouse.click`, and `page.mouse.click` resolves when the event is dispatched, not when the page has handled it. `importAppend` is several promise hops long. It passes today because the `openDialogCount` round trip in between gives the microtasks somewhere to land. `expect.poll` would make that intentional instead of lucky. Same shape as the wheel note in `zoom-ladder.spec.ts`.

## Where it goes from here

The panel is a good addition and the icon research behind it is the right kind of work. Merge base first, since that answers the geometry question I raised last time without any code. The import truncation is the one I would want fixed before it lands, because it fails silently on real files. Everything else here can follow.

wormeyman:
Re-reviewed at `c107fbd0`. Seven of the eight items from last time are fixed, and I drove the branch rather than only reading it, since two of them were geometry and one was a silent truncation. Every number below was measured at `c107fbd0` with both dev servers up.

## Fixed, and checked

**The import truncation is gone, end to end.**

```
source string length : 2417093
maxlength attribute  : null       (was 1048576)
after a real paste   : 2417093    all of it
Replace -> entities  : 1636       no page errors
```

Control in the same run: I put `maxLength = 2 ** 20` back on the live element and pasted again, and got exactly 1048576. So the attribute was the cause and dropping it is the fix.

**The clamp works.** Six widths, with the export-image slot's right edge measured against the viewport:

| viewport | panel   | last slot's right edge |
| -------- | ------- | ---------------------- |
| 1280     | 861..1073 | 1061                 |
| 1000     | 721..933  | 921                  |
| 900      | 671..883  | 871                  |
| 866      | 654..866  | 854                  |
| 800      | 588..800  | 788                  |
| 700      | 488..700  | 688                  |

Nothing off screen at any of them.

**`safeIcon` holds.** I mutation-checked it by pointing the redo slot at a name FD does not have:

|                  | editor comes up | what the user gets              |
| ---------------- | --------------- | ------------------------------- |
| with `safeIcon`  | yes             | one blank slot, toast naming it |
| bypassing it     | no              | `UNRECOVERABLE_ERROR`, blank page |

The other four I checked by reading: `generateSlots` clears before it places, the ticker comes off on `destroy`, the panel sizes itself from `cells.length`, and the wires read copper, red, green again. CI is green on this head, all four Playwright shards included.

## What is left

### Merging base is clean, and it settles the toast question

Still six commits behind. I merged base into a throwaway copy of the branch: no conflicts, `vp check` clean, `vp test` 184 passing across 14 files. #233's new spec guard finds nothing to flag in `quick-actions.spec.ts`, so that one costs you nothing.

With base merged, three live toasts up, `document.elementFromPoint` at each slot centre:

| slot     | hit      |
| -------- | -------- |
| all nine | `CANVAS` |

Two things follow. The comment on `ROWS` still argues that the two-row wrap is what keeps the panel out from under `.toasts-container`. After the merge that stops being true, and leaving it will send the next reader hunting for a constraint that is gone. And the buttons are still *drawn under* those toasts for the first 30 seconds after a load: the container measures x 960..1280, and five of the nine slot centres sit at 967 or further right. A click getting through is not the same as being able to see what you are clicking. That is the layout question from last time, still open.

### `safeIcon` covers five of the eight icons that can throw

Five icons go through `safeIcon`. The three wire slots do not: `WireSlot`'s own constructor calls `F.CreateIcon(wireName)`, and `buildCells` only writes `new WireSlot(WIRES[0])`, so the guard never sees it. The ninth slot, ALT, builds a pixi `Text` and reads no FD, so it cannot throw either way. Same measurement as above, with `green-wire` swapped for a name FD lacks:

```
editor ready : false
page error   : Error: UNRECOVERABLE_ERROR
```

That is the exact failure `safeIcon` was added to stop, still live for a third of the panel. Nothing reaches it today, because the wire names are a hardcoded const, but the same was true of the five that are now guarded.

### The clamp trades one problem for another, and the comment reads the other way round

Below a threshold the panel now sits on top of the quickbar. Measured:

| viewport | overlap | quickbar slots covered                       |
| -------- | ------- | -------------------------------------------- |
| 800      | 33px    | none, it lands in the panel's 12px padding    |
| 778      | 55px    | column 9, so 2 slots                          |
| 702      | 119px   | columns 8 and 9, so 4 slots                   |

At 700x800 I clicked quickbar slot 9 and got the Import dialog. The quickbar never saw it.

I am not treating this as a blocker. It starts at 778 where the bug it replaces started at 866, and it is a narrow band either way. But the comment on `setPosition` says the overlap "is the same trade-off a real user can still click through", which reads as though the quickbar keeps working underneath, and it does not. Reword it, or lift the panel above the quickbar below 778.

### Neither fix is pinned

Nothing under `tests/` reads `maxLength` or the panel's position, so the cap could come back and the clamp could be deleted with every check green. Both of them fail quietly, which is the shape worth pinning. `quick-actions.spec.ts` already drives the import textarea, so the cap one is a few lines: insert something over a megabyte, assert all of it arrived.

## Two small ones

`packages/website/src/index.ts:342` still calls `exportString` one of "the other two QuickActions members". It is not one any more.

`safeIcon`'s doc says the fallback is "a blank square the same size as a real icon". It returns an empty `Container`, which draws nothing. The slot still renders and still takes a click, which I measured, so the behaviour is right and only the sentence is wrong.

## Where it goes from here

Two things and I will merge it. Merge base in, which is free: I tried it and there are no conflicts, `vp check` is clean and `vp test` is 184 passing. And wrap the three `WireSlot` icons in `safeIcon` the same way the other five are, which is about three lines and closes the last third of the case that commit was written for.

The rest can follow: the wording on `ROWS` and `setPosition`, the two doc lines at the end, and a test for each fix. I would rather have those than not, but none of them should hold the panel up.

wormeyman:
I have detached this repo and it is standalone now you may have to re-fork this repo and resubmit the PR's. @koenigstag

wormeyman:
reopening

PR #227, opened 2026-08-13
Draft: Add a BookDialog tree view, opened from a button beside BlueprintInfoButton

koenigstag:
The only way to switch entries in a loaded blueprint book was a bare "BP Book Index" number field in the settings pane, with no names, icons, or indication of nesting.

BookDialog walks the book's raw entries directly (not the flattened index space `Book.selectBlueprint` reads) and renders them as an always-expanded tree — nested books get a header row and indented children, planners get a dimmed placeholder row, and a depth guard of 10 stops runaway recursion on a malformed book. Rows scroll with the same mask+wheel+thumb pattern `InventoryDialog` already uses.

`BookButton` opens it, sitting one slot pitch right of `BlueprintInfoButton` and visible only while a book is loaded, rather than living in ToolsPanel (#221).

Stacked on #222 — future base is `wormeyman-space-age-support`, not the default branch.

image

--

image

### DIFF
--- a/packages/editor/src/Editor.ts
+++ b/packages/editor/src/Editor.ts
@@ -11,7 +11,7 @@ import { Application, TextureSource, setBasisTranscoderPath, Assets } from 'pixi
 import basisTranscoderJS from './basis/transcoder.1.16.4.js?url'
 import basisTranscoderWASM from './basis/transcoder.1.16.4.wasm?url'
 import { loadData } from './core/factorioData'
-import G, { Logger } from './common/globals'
+import G, { Logger, QuickActions } from './common/globals'
 import { Entity } from './core/Entity'
 import { Blueprint, oilOutpostSettings, IOilOutpostSettings } from './core/Blueprint'
 import { BlueprintContainer, EditorMode, GridPattern } from './containers/BlueprintContainer'
@@ -21,16 +21,28 @@ import { Dialog } from './UI/controls/Dialog'
 import { ActionRegistry, MouseButton } from './actions'
 import { IPoint } from './types'
 
+/**
+ * A single object rather than trailing positional parameters, so a required
+ * `quickActions` and an optional `logger` don't have to fight over which
+ * comes first, and a third option later doesn't reshuffle anyone's call site.
+ */
+export interface EditorInitOptions {
+    quickActions: QuickActions
+    logger?: Logger
+}
+
 export class Editor {
-    public async init(canvas: HTMLCanvasElement, logger?: Logger): Promise<void> {
+    public async init(canvas: HTMLCanvasElement, options: EditorInitOptions): Promise<void> {
         setBasisTranscoderPath({ jsUrl: basisTranscoderJS, wasmUrl: basisTranscoderWASM })
 
         TextureSource.defaultOptions.scaleMode = 'linear'
         TextureSource.defaultOptions.addressMode = 'repeat'
 
-        if (logger) {
-            G.logger = logger
+        if (options.logger) {
+            G.logger = options.logger
         }
+
+        G.quickActions = options.quickActions
 
         const app = new Application()
 
--- a/packages/editor/src/UI/BookButton.ts
+++ b/packages/editor/src/UI/BookButton.ts
@@ -1,0 +1,48 @@
+import { Container, FederatedPointerEvent } from 'pixi.js'
+import G from '../common/globals'
+import { Button } from './controls/Button'
+import F from './controls/functions'
+
+/**
+ * Persistent top-left button that opens BookDialog.
+ *
+ * Positioned just right of the website's own DOM corner overlay (the FBE logo
+ * plus the Discord/Github buttons in packages/website/index.html), which is
+ * 140px wide and sits at the fixed (0, 0) corner - a canvas-drawn button at
+ * that x would render underneath it. Left rather than right avoids
+ * EntityInfoPanel instead, which docks flush against the right edge whenever
+ * an entity is hovered. A standalone corner button rather than a ToolsPanel
+ * slot, since "is a book loaded" is orthogonal to which quick-action panel
+ * happens to exist.
+ *
+ * Visible only while a book is loaded - polls `G.quickActions.getCurrentBook()`
+ * from the ticker rather than checking once, since "is a book loaded" changes
+ * at runtime (a book replaced by a bare blueprint, or the reverse) with
+ * nothing here to be re-constructed and notice it.
+ */
+export class BookButton extends Container {
+    private readonly m_Button: Button<undefined>
+
+    public constructor() {
+        super()
+
+        this.m_Button = new Button<undefined>(undefined, 36, 36)
+        this.m_Button.content = F.CreateIcon('blueprint-book', 24)
+        this.m_Button.on('pointerdown', this.onPointerDown)
+        this.addChild(this.m_Button)
+
+        this.position.set(152, 6)
+        this.visible = false
+
+        G.app.ticker.add(() => {
+            this.visible = G.quickActions.getCurrentBook() !== undefined
+        })
+    }
+
+    private readonly onPointerDown = (e: FederatedPointerEvent): void => {
+        e.stopPropagation()
+        if (e.button === 0) {
+            G.UI.toggleBookDialog()
+        }
+    }
+}
--- a/packages/editor/src/UI/BookDialog.ts
+++ b/packages/editor/src/UI/BookDialog.ts
@@ -1,0 +1,266 @@
+import { Container, Graphics, Rectangle, Text } from 'pixi.js'
+import { Book } from '../core/Book'
+import { IBlueprintBookEntry } from '../types'
+import G from '../common/globals'
+import F from './controls/functions'
+import { Button } from './controls/Button'
+import { Dialog } from './controls/Dialog'
+import { styles } from './style'
+
+const WIDTH = 320
+const PADDING = 12
+const ROW_HEIGHT = 24
+const ICON_SIZE = 18
+const INDENT = 14
+const VP_X = PADDING
+const VP_Y = 40
+const VP_W = WIDTH - PADDING * 2
+const VP_H = 320
+const HEIGHT = VP_Y + VP_H + PADDING
+
+/**
+ * How many `blueprint_book`s deep to descend. A guard, not a real limit -
+ * nothing seen in the wild nests anywhere close to this - so hitting it
+ * shows one placeholder row rather than silently dropping the rest of the
+ * tree, the same way a corrupt or hand-crafted string could.
+ */
+const MAX_DEPTH = 10
+
+const ROW_LABEL_STYLE = styles.dialog.label
+const ACTIVE_ROW_COLOR = 0xb16925
+
+/** A book's own icon, or its first blueprint icon, or nothing - every one
+ * of those is optional in the format, and a bad signal name in an icon
+ * throws, which is worth losing one icon over rather than the whole dialog. */
+function tryCreateIcon(name: string | undefined, size: number): Container | undefined {
+    if (name === undefined) return undefined
+    try {
+        // setAnchor stays at its default (true/centred) - every call site
+        // positions the icon by its centre.
+        return F.CreateIcon(name, size)
+    } catch {
+        return undefined
+    }
+}
+
+function firstIconName(
+    icons: readonly { signal: { name?: string } }[] | undefined
+): string | undefined {
+    return icons?.[0]?.signal.name
+}
+
+/** A non-interactive row - a nested-book header, a planner placeholder, or
+ * the depth-limit notice - dimmed to read as informational rather than
+ * clickable, since nothing here reacts to a pointer. */
+function createStaticRow(text: string, x: number, y: number, iconName?: string): Container {
+    const row = new Container()
+    row.position.set(x, y)
+    row.alpha = 0.6
+
+    const icon = tryCreateIcon(iconName, ICON_SIZE)
+    let labelX = 0
+    if (icon) {
+        icon.position.set(ICON_SIZE / 2, ROW_HEIGHT / 2)
+        row.addChild(icon)
+        labelX = ICON_SIZE + 4
+    }
+
+    const label = new Text({ text, style: ROW_LABEL_STYLE })
+    label.position.set(labelX, (ROW_HEIGHT - label.height) / 2)
+    row.addChild(label)
+
+    return row
+}
+
+/** A clickable row for one selectable blueprint entry. */
+function createBlueprintRow(
+    label: string,
+    iconName: string | undefined,
+    x: number,
+    y: number,
+    width: number,
+    active: boolean,
+    onSelect: () => void
+): Container {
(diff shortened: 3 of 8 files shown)


wormeyman:
Still a draft, so this is orientation rather than a review. Two things will bite at merge and are much cheaper to know now than after more work goes on top.

## The button coordinate collides with #222

`BookButton` sits at `position.set(152, 6)`. That is the same coordinate as `BlueprintInfoButton` in #222, and this branch does not contain #222, so if both land the two buttons sit exactly on top of each other. The class comment says the button goes beside `BlueprintInfoButton`, which is what it should do but not what the code does. Whichever of the two rebases second has to pick a real second slot.

## `QuickActions` is being defined twice

#221 also adds a `QuickActions` interface to `common/globals.ts`, and also reshapes `Editor.init` into an `EditorInitOptions` object. The members do not overlap at all. #221 has `importReplace`, `importAppend`, `exportString`, `exportImage`, `encodeCurrent` and `readClipboardText`; this branch has `getCurrentBook` and `selectBookEntry`.

Git will flag the conflict, but resolving it is a design call rather than a merge. It has to end up as one interface carrying both sets, not whichever version lands second. Rebasing on #221 before going further would mean you are adding two members to an existing interface instead of writing a second one and then reconciling.

## One small thing

`Book.entries` hands back the live `blueprints` array. The `readonly` return type stops honest callers and nothing else, so a cast reaches straight into the book's own state. `BookDialog` only reads it, so this is about the next caller rather than this one.

The tree view itself reads well. Threading `flatIndex` through the recursion so a row's index means what `selectBlueprint` means by it is the right way round, and `tests/book-serialize.spec.ts` exists because that index space is easy to get wrong, so it is worth a look while you are in there.

wormeyman:
Still a draft, and unchanged since my orientation comment, so this is not a merge review. I read it properly this time while reviewing #221 and #222 together, and four things are worth having before more work goes on top. Two of them are structural and cheap to fix now.

Everything in my earlier comment still stands: the `(152, 6)` collision with #222's button, the two competing `QuickActions` definitions, and `Book.entries` handing out the live array.

## The scroll viewport's hit area scrolls with the content

This is the one to fix first, because it makes the list stop working rather than look wrong.

`InventoryDialog` keeps `m_InventoryItems` parked at `(VP_X, VP_Y)`, puts the hit area on it, and scrolls an inner child. You can see it at line 312, where the scroll amount is read as `-active.y`, off the group container rather than off the one holding the hit area.

`BookDialog` collapses those two into one. `m_Rows` carries the hit area at line 208 and is also what the wheel handler moves at line 239. The rectangle is in `m_Rows`' own space, so it travels with it. Scroll down by S pixels and the live region is only the top `VP_H - S` of the viewport: the bottom S pixels take neither a click nor a wheel event. A book with 27 or more blueprints has `maxScroll >= VP_H`, so at full scroll the hit area sits entirely above the viewport and the list goes completely dead, with no way to scroll back.

There is a smaller version of the same thing even at rest. The rectangle starts at local x = 0 while every row spans x 12 to 308, so 12 pixels of padding are live and the right-hand 12 pixels of every row are not.

Splitting it back into a fixed outer container and a scrolling inner one is the whole fix, and it is what the file it was copied from already does.

## The active-row highlight never shows

`Button`'s constructor ends with `this.addChild(this.m_Background, this.m_Active, this.m_Hover)`, so `m_Background` is child 0. `button.addChildAt(highlight, 0)` at line 89 puts the highlight in front of that in the child list, which means behind it on screen. `colors.controls.button.background` is `{color: 0x646464, alpha: 1}`, fully opaque, so nothing of the highlight survives. Open a book on any entry and no row is marked.

The part that makes this worth a second look is that `Button.m_Active` already draws this. `colors.controls.button.active` is `{color: 0xb16925, alpha: 1}`, and `ACTIVE_ROW_COLOR` at line 30 is `0xb16925`. So the hand-rolled graphic is both invisible and a second copy of a mechanism that is already there. `button.active = true` is the whole thing.

## The depth guard makes later rows point at the wrong blueprint

When `depth >= MAX_DEPTH` the code pushes a placeholder row and returns without recursing, so the blueprints inside that subtree never advance `flatIndex.current`. `Book.getBlueprintAtFlattenedActiveIndex` still counts them.

Given entries `[a book nested too deep holding 3 blueprints, bpA]`, bpA is drawn carrying index 0, and clicking it calls `selectBookEntry(0)`, which loads the first blueprint inside the skipped book. The highlight lands on the wrong row for the same reason.

Walking the subtree for its count even when not drawing it is the fix. Worth noting that the mutable `{ current: number }` box is what let this happen quietly: a returned index would have made the missing increment a visible hole in the control flow.

## `firstIconName` and `BookButton` both need the guard this file already has

Two spots read an icon without the protection `tryCreateIcon` provides three lines away.

`icons?.[0]?.signal.name` at line 49 guards `icons` and `icons[0]` but not `signal`. The schema marks `signal` required, but validation is lenient on purpose, so a blueprint carrying `icons: [{index: 1}]` loads and then throws inside `buildRows`, outside `tryCreateIcon`, with nothing catching above `new BookDialog(book)`. One `?.` closes it.

`BookButton` calls `F.CreateIcon('blueprint-book')` in `UIContainer`'s constructor, which `Editor.init` calls with nothing catching above it. That is not hypothetical: #222 has a live crash of exactly this shape. `F.CreateIcon` throws on any name outside items, fluids, recipes, signals and inventory groups, and I measured 17 icon references in the committed corpus that land there. `tryCreateIcon` in this file is exactly the guard #222 is missing, so it would be worth putting one copy somewhere both branches can use.

## Smaller

`BookButton` polls `G.quickActions.getCurrentBook()` from the render ticker every frame, forever, with no `ticker.remove`, for state that only changes when `loadBp` runs. It is also constructed on mobile, where `UIContainer` never adds it, so the poll runs for the life of the page writing `.visible` on a container with no parent.

`BookButton` at (152, 6) also overlaps `DebugContainer`, which sets `x = 145; y = 5` and draws its FPS line at its own origin. The button is added after it, so with debug on the readout goes behind the button. The comment justifying 152 reasons about the DOM `#corner-panel` and misses the canvas-side neighbour.

Every row is built eagerly, with three sprites and a `Text` whose height is measured on the spot, though only about 13 fit in the viewport and the dialog is rebuilt from scratch on every open.

An entry that is none of blueprint, book, upgrade planner or deconstruction planner produces no row at all. The `MAX_DEPTH` comment says the point of a placeholder is to avoid "silently dropping the rest of the tree", and a trailing `else` would hold to that for one line.

`tryCreateIcon` catches with a bare `catch {}` and no log, so a missing icon leaves nothing to identify which name failed. `getSpriteData` and `OverlayContainer.createEntityInfo` both turn the same failure into a named line.

## On tests

There are none, and the flat-index walk in `buildRows` is pure and needs no `FD`, which puts it in the same class as `railSignalSnapping.test.ts` and `zoomLevels.test.ts`: unit testable under `vp test`, seconds in CI, no browser. Given item 3 above is arithmetic that goes wrong silently, that is where I would start rather than with a Playwright spec. `tests/book-serialize.spec.ts` exists because this index space is easy to get backwards, and it is worth reading before the fix rather than after.

The tree view itself still reads well, and threading `flatIndex` so a row's index means what `selectBlueprint` means by it is the right call. Rebasing on #221 before going further would also turn the `QuickActions` collision into two added members rather than a merge to resolve.

wormeyman:
Reviewed `ebe28b3d` while it is still a draft, because two of these are cheaper
to know before the rebase than after.

Measured against base `823400f4`. GitHub reports `mergeable: CONFLICTING`, and
`git merge-tree` conflicts in 4 of the 6 files: `UIContainer.ts` (5 regions),
`Editor.ts` (2), `packages/website/src/index.ts` (2), `globals.ts` (1). Merging
#243 first makes it worse, not better.

## The wiring needs a rewrite, not a conflict resolution

The whole layer this PR introduces already landed on base under the same two
names, with disjoint contents. Base commit `9aee2ac7` added its own
`QuickActions`:

| | base | this PR |
| --- | --- | --- |
| members | `importReplace`, `importAppend`, `exportImage`, `encodeCurrent`, `readClipboardText` | `getCurrentBook`, `selectBookEntry` |

Base also already changed `init(canvas, logger?)` to take an options object,
including `G.quickActions = options.quickActions ?? noopQuickActions`.

What makes this worth flagging before the rebase is that the collision is
partly invisible. Both sides added the byte-identical
`.init(CANVAS, { quickActions, logger: createToast })` line and the
byte-identical `export type { QuickActions }` to the barrel, so those hunks
merge with no conflict around a type whose definition is whichever arm of the
`globals.ts` conflict wins.

**And one arm merges cleanly and silently kills the feature.** Line 56,
`addChild(this.quickbarPanel, this.wiresPanel, this.bookButton)`, conflicts with
base's `addChild(this.quickbarPanel, this.toolsPanel)`. `WiresPanel` no longer
exists on base, so a resolver must take base's arm - and the natural move is to
take base's arm on the field declaration too. The result: `BookButton` is
constructed, its ticker callback runs at ~60 Hz for the life of the page, and
the button is never added to the display list. `toggleBookDialog` compiles as
dead code. `vp check` passes, all four shards pass, and there is no book button
on screen.

Concretely, the rebase is smaller than the diff suggests:

1. Drop this PR's `QuickActions` and `EditorInitOptions` entirely. Add
   `getCurrentBook` and `selectBookEntry` as two members of base's existing
   interface.
2. Keep base's optional `quickActions?:` and its `?? noopQuickActions`. Add
   matching members to `noopQuickActions` - but have `getCurrentBook` return
   `undefined` **silently**, unlike every sibling. `BookButton` polls it from
   the ticker every frame, and `G.logger` is a toast, so the usual
   report-and-degrade pattern would produce ~60 toasts a second.
3. Add `this.bookButton` to base's `addChild(...)` line by hand.
4. Do not touch the `import.meta.env.DEV` block in
   `packages/website/src/index.ts`. This branch predates #302 and assigns
   `window.__fbe_test` unconditionally; base's guard is what keeps the ~400-line
   test API out of production. The merge keeps base's version, but a rebase that
   takes this file wholesale would put it back with every CI signal green.
5. `F.CreateIcon('blueprint-book', 24)` at `BookButton.ts:30` should be
   `F.SafeIcon`. It runs in `UIContainer`'s constructor inside `Editor.init`,
   where a throw fails the whole init. `SafeIcon` did not exist when this
   branched; `BookDialog`'s own `tryCreateIcon` already gets this right.

The `getCurrentBook` / `selectBookIndex` hoist is worth keeping. I compared
base's `changeBookIndex` against base's `testApi.selectBookIndex` and the bodies
are identical, so deduping them is behaviour-neutral.

## Three bugs in the dialog itself

**1. The hit area slides with the scroll, so rows go dead.** `m_Rows` is given
`hitArea = new Rectangle(0, 0, VP_W, VP_H)` and is also the object the wheel
handler moves (`this.m_Rows.y = ...`). A `hitArea` is in local space, so at
scroll `s` the interactive band is `[VP_Y - s, VP_Y - s + VP_H]` while the
visible band stays put. One wheel notch on a 20-entry book makes the bottom
visible row unclickable; past 320 px of scroll nothing in the list is clickable
and the wheel stops working, with no way to scroll back.

`InventoryDialog` does not do this - it fixes the container at `(VP_X, VP_Y)`
and scrolls a *child* (`active.y`). This one sets `(0, VP_Y)` and scrolls the
container. The `x = 0` rather than `VP_X` separately kills the right 12 px of
every row.

**2. The active-row highlight never renders.** `button.addChildAt(highlight, 0)`
puts it at index 0, but `Button`'s constructor has already run
`this.addChild(this.m_Background, this.m_Active, this.m_Hover)`, and
`m_Background` is a filled rect at alpha 1. The orange bar is drawn and painted
over every time, so nothing marks the currently-open blueprint - which is half
of what this PR exists to fix. `Button` already owns this: `m_Active` is
`colors.controls.button.active`, the same `0xb16925`, so `button.active = active`
is the fix.

**3. Past the depth guard, clicking a row selects a different blueprint.** In
ordinary books the mapping is correct - I checked nested books, a planner
mid-list, an empty nested book and a book as the first entry, and every row
resolves to the blueprint whose label it shows. The exception is the guard
itself: at `depth >= MAX_DEPTH` it pushes a placeholder without recursing, so
`flatIndex` never advances past the hidden blueprints while `Book`'s flattened
space still counts them. Every row after a too-deep book is off by that count. A
root book of `[bp "H0", <11 levels of nesting holding "Hdeep">, bp "H1"]` makes
the row labelled `H1` select `Hdeep`.

Needs 11 levels of nesting, so not something Factorio emits - but
`blueprint_book` is recursive in `blueprintSchema.json` with no depth cap, and
`bpString.ts` loads whether validation passes or fails. The file comment says
hitting the guard "shows one placeholder row rather than silently dropping the
rest of the tree"; it also silently corrupts every row below, which is worth
saying there. Cycles are not a concern - the book comes from `JSON.parse`, which
cannot express one.

Minor, same file: `firstIconName` does `icons?.[0]?.signal.name`, and
`definitions.signal` in the schema is `["object", "null"]`. A null signal throws
from the constructor, outside `tryCreateIcon`'s try, so the dialog fails to open
rather than losing one icon. `signal?.name` costs nothing.

Credit where due: the icons **are** drawn through a try boundary, which is the
right handling of the `F.CreateIcon` planet-name limit, and this is the first
place in the codebase that draws blueprint icons at all. I also checked the
"book swapped while the dialog is open" case and it is not reachable, because
every path routes through `Editor.loadBlueprint`, which calls `Dialog.closeAll()`
synchronously.

## Two smaller things

**The button position collides with #243.** `BookButton.ts:34` is
`this.position.set(152, 6)` at 36x36. #243's `BlueprintInfoButton` is
`this.position.set(152, 6)` at 36x36. Identical pixel, identical size, same
`addChild` call, and pixi hit-tests the later child first - so whenever a book is
loaded one of the two becomes unclickable. Worth picking a real offset during
the rebase, and #243 could export a named pitch rather than both files carrying
the literal.

**The PR body is stale in two ways.** `git merge-base --is-ancestor` says #227's
head does not contain #243's commits; `git log cfac9b03..ebe28b3d` is exactly one
commit on a plain base-branch parent. So there is no stack in git, and
`git grep BlueprintInfoButton` at this head returns nothing - the symbol exists
in neither tree. The real dependency is the reverse of what the body says, and
it is only the pixel above.

## Before it leaves draft

Nothing in the four Playwright shards can reach a single line of `BookDialog` -
there is no spec and no test hook that opens it, so the green CI on this PR says
nothing about the feature. Findings 1 and 2 are both "suite green, feature
broken" cases. A minimum set, modelled on the existing `tools-panel.spec.ts` and
`settings-pane-book-index.spec.ts` which already have the hooks:

- click the button, assert `openDialogCount()` goes 0 to 1 - this is the one that
  catches a BookButton that was never added to the display list;
- a control on a bare blueprint: same click, count stays 0, no page error;
- with `[bp, book[bp, bp], planner, bp]`, click the last row and assert
  `activeIndex === 3`, pinning that the two walks agree.

And per the repo's own rule for anything with a threshold or a step order: drive
the scroll by hand on a long book and print the numbers, rather than trusting a
green suite.

PR #242, opened 2026-08-19
Add a ToolsPanel with entity-info toggle/import string/export string/export image/undo/redo UI actions

koenigstag:
Re-submission of #221, recreated after the repo detach.

## Summary
- Expose import-replace, import-append, export-to-string and export-to-image — actions that only ever existed as keyboard shortcuts (paste, Ctrl+Shift+V, copy, Ctrl+S) — as buttons, plus Undo/Redo (Ctrl+Z/Ctrl+Y).
- Added ImportDialog (textarea + Replace/Append buttons with descriptions) and ExportDialog (pre-filled, pre-selected textarea) as a clipboard-free alternative, for browsers that block `navigator.clipboard` or a string that needs to be seen/edited.
- Renamed WiresPanel to ToolsPanel, since it now holds import/export/undo/redo alongside the wire slots.

image

image

## Layout notes
- The panel wraps into a column-major grid instead of a single wide row — a wide single row runs part of it under `.toasts-container` (`position: fixed; right: 0`), confirmed by clicking a widened-panel slot and finding the click reached the toast instead of the button.
- Undo/Redo call `G.bp.history.undo()/redo()` directly (same as their keybinds in Editor.ts); the four clipboard/file actions go through a new `QuickActions` interface injected via `Editor.init()`, mirroring the existing `Logger` injection.
- Icons: game GUI sprites (`import_slot`/`export_slot`) and the `signal-anticlockwise-circle-arrow`/`signal-clockwise-circle-arrow` virtual signals for Undo/Redo — the only undo/redo-shaped icons anywhere in vanilla Factorio's data.

## Test plan
- [x] `vp check` / `vp test` pass
- [ ] Manually verified all quick actions, both dialogs, and Undo/Redo end-to-end against the dev server


koenigstag:
PR recreated after the repo detach (was #221). Ready for review — waiting on any fixes/comments.

packages/editor/src/UI/ExportDialog.ts
```diff
@@ -0,0 +1,49 @@
...
+        )
+        this.m_TextInput.placeholder = 'The current blueprint is empty.'
+        this.m_TextInput.position.set(PADDING, FIELD_Y)
+        this.addChild(this.m_TextInput)
+
+        G.quickActions
+            .encodeCurrent()
+            .then(source => {
+                this.m_TextInput.text = source ?? ''
+                this.m_TextInput.select()
```
wormeyman: **The pre-selection never happens, so this dialog falls back to the clipboard API it exists to avoid.**

`TextInput._onAdded` (`TextInput.ts:295`) sets `style.display = 'none'` on `addChild`, which fires inside this constructor at line 37. `encode()` (`bpString.ts:249`) resolves *synchronously* inside its Promise executor, so this `.then` runs in a microtask - before any frame renders, and so before `_updateDOMInput` -> `_setDOMInputVisible` ever sets `display: block`. The `text` setter one line up only writes `_dom_input.value`; it doesn't trigger an update either.

`focus()` on a non-rendered element is a no-op, so `document.activeElement` stays the `CANVAS`. The user presses Ctrl+C, the document `copy` listener sees `activeElement === CANVAS` and routes to `exportString()` -> `navigator.clipboard.writeText`. In the target case - a browser that blocks the clipboard - they get an error toast and have to hand-select a 2.4 MB textarea.

Defer the `select()` until after a render tick (or force an update before it), rather than relying on the promise being async.

packages/editor/src/UI/ExportDialog.ts
```diff
@@ -0,0 +1,49 @@
...
+ * silently placed on the clipboard. Pre-filled and pre-selected on open, so
+ * a plain Ctrl/Cmd+C is enough once the dialog is up.
+ */
+export class ExportDialog extends Dialog {
+    private readonly m_TextInput: TextInput
+
+    public constructor() {
+        super(WIDTH, HEIGHT, 'Export')
+
+        this.m_TextInput = new TextInput(
```
wormeyman, line 64: **This field is freely editable and encoded exactly once, so it can silently show a stale string.**

The dialog doesn't block the canvas. Open Export on a loaded blueprint, leave it open, place or delete an entity behind it, then copy the field - you get the pre-edit string, with nothing indicating it's stale.

Typing into it has the same problem from the other side: the sibling `ImportDialog`'s identically styled field *does* respond to editing, so this one looks like it should too, and the text is just discarded on close.

`TextInput` already has a `disabled` setter that sets `_dom_input.disabled` - though note it also flips the box to the `DISABLED` state, so `readOnly` on `htmlInput` may read better here. Re-encoding on a render tick would fix the staleness half as well.

packages/editor/src/UI/ImportDialog.ts
```diff
@@ -0,0 +1,106 @@
...
+ * other keeps it - so each of those three gets its own line of description
+ * rather than trusting a button label alone to carry the distinction.
+ */
+export class ImportDialog extends Dialog {
+    private readonly m_TextInput: TextInput
+
+    public constructor() {
+        super(WIDTH, HEIGHT, 'Import')
+
+        this.m_TextInput = new TextInput(
```
wormeyman, line 53: **Escape can't close this dialog once the textarea has focus, and there's no close control to fall back on.**

`Editor.ts:767` drops the keydown before `G.actions.pressKey`:

```ts
if (e.target instanceof HTMLTextAreaElement) return
```

so the `closeWindow` action and `Dialog.closeLast()` never run. `Dialog` draws no close button and `UIContainer` registers no outside-click handler, so the only exits are clicking the ToolsPanel slot again or clicking the canvas first to blur.

This is on the dialog's primary flow - paste a string, change your mind - and it applies to `ExportDialog` too. A keydown listener on the textarea that closes on Escape would cover both.

packages/editor/src/UI/ImportDialog.ts
```diff
@@ -0,0 +1,106 @@
...
+        addDescribedButton(
+            this,
+            PADDING,
+            REPLACE_Y,
+            WIDTH,
+            'Replace',
+            'Replaces the whole blueprint with the one above.',
+            () => {
+                G.quickActions.importReplace(this.m_TextInput.text)
+                this.close()
```
wormeyman: **`close()` runs synchronously, before the async import can fail, so a bad string takes the user's text with it.**

`importReplace(text)` starts an async chain; this destroys the dialog and its textarea immediately; then `getBlueprintOrBookFromSource` rejects and a toast appears. The hand-edited string is gone and has to be retyped - which defeats the case the class doc names ("a string being hand-edited rather than pasted whole").

The empty-field case is worse: `importReplace('')` calls `loadingScreen.show()`, then `getBlueprintOrBookFromSource('')` falls past the `DATA[0] === '0'` branch and builds `new URL('https://')`, which throws - so an empty field reports **"Invalid URL"** rather than "paste a blueprint string".

Closing on success only (and guarding the empty string up front) would fix both. Same applies to Append below.

packages/editor/src/UI/ToolsPanel.ts
```diff
@@ -0,0 +1,303 @@
...
+ * rather than the momentary hover/press one `Button` already draws. Used for
+ * Alt, so the button shows whether entity info is visible even when the
+ * state last changed through the `AltLeft` keybind rather than a click here.
+ * Inserted just below the slot's own content (added last, so it's the final
+ * child) so the icon/text stays legible on top of it.
+ */
+function addToggleHighlight(slot: Container, color: number): Graphics {
+    const highlight = new Graphics().rect(0, 0, 36, 36).fill(color)
+    highlight.visible = false
+    slot.addChildAt(highlight, slot.children.length - 1)
```
wormeyman: **This inserts the highlight *above* the hover and active sprites, not just below the content - and it's fully opaque.**

`Button` adds `[m_Background, m_Active, m_Hover]` (`Button.ts:75`) and the `content` setter appends `m_Content`, so a constructed `Slot` has four children: `[background, active, hover, content]`. `children.length - 1` = 3 puts the highlight at index 3 - above `m_Hover` (2) and `m_Active` (1).

`new Graphics().rect(0, 0, 36, 36).fill(0xf1be64)` has no alpha, so while Alt is on, the button gives no hover and no press feedback and loses its raised border. That's the one slot that most needs to keep looking pressable, since it's the only toggle in the panel.

Two smaller things at the same site: the 36x36 is a third copy of `Slot`'s default size rather than being read from the slot; and the signature takes `slot: Container`, so a container with zero children throws `The index -1 supplied is out of bounds` - nothing in the type or the doc rules that out.

packages/editor/src/UI/ToolsPanel.ts
```diff
@@ -0,0 +1,303 @@
...
+
+/**
+ * An icon that failed to build costs this one slot rather than the whole
+ * panel - `F.CreateIcon`/`F.CreateUtilitySpriteIcon` throw by design for a
+ * name FD does not have (see `need()` in `core/need.ts`), and nothing above
+ * `generateSlots()` catches. Falls back to a blank square the same size as a
+ * real icon, so the slot still exists and is still clickable; only the icon
+ * itself is missing, named in a warning rather than silently swallowed.
+ */
+function safeIcon(name: string, build: () => Container): Container {
```
wormeyman, line 148: **The doc promises a visible fallback that the code doesn't draw.**

> Falls back to a blank square the same size as a real icon, so the slot still exists and is still clickable

It returns `new Container()` - no children, no bounds, nothing rendered. If `FD.utilitySprites.import_slot` ever goes missing (a `data.json` regenerated against a different Factorio version), the Import slot renders as an empty square indistinguishable from an unassigned quickbar slot, with only a toast to say why.

Either draw the promised placeholder (a `Graphics` rect at the slot's size would do it) or drop the claim from the comment.

packages/editor/src/UI/ToolsPanel.ts
```diff
@@ -0,0 +1,303 @@
...
+    /**
+     * Re-callable: clears whatever `slotsContainer` currently holds (and
+     * destroys it - `removeChildren()` alone only detaches, it does not
+     * release the GPU resources a re-generated panel would otherwise leak)
+     * before placing a freshly built set, and replaces the ticker rather than
+     * accumulating a second one. Nothing calls this a second time today, but
+     * nothing should have to trust that either - `QuickbarPanel.generateSlots`
+     * is the precedent this mirrors, for row-count changes.
+     */
+    public generateSlots(): void {
```
wormeyman, line 278: **This is unreachable, and it couldn't do the job its doc describes if it were called.**

Zero call sites. Unlike `QuickbarPanel.generateSlots(items)` - the precedent the doc cites - it takes no arguments, so every call rebuilds an identical nine-cell grid. And if the cell list ever *did* change length it would still be broken: the panel background is sized from `cols` in the constructor and `placeCells` never resizes it, so an extra column would be drawn outside the panel.

One factual correction in the comment: it says `removeChildren()` "does not release the GPU resources" that `child.destroy()` does. `Container.destroy()` with no options doesn't destroy children either - the slot's background/active/hover sprites survive exactly as they would have - and those sprites share cached `rect_map` RenderTextures anyway, so there's nothing per-slot to release.

Suggest deleting it until something needs it.

packages/editor/src/UI/ToolsPanel.ts
```diff
@@ -0,0 +1,303 @@
...
+     * Flush against the quickbar's right edge at common viewport widths, the
+     * same way the two-row layout above assumes - but clamped to the screen's
+     * own right edge underneath that, since the unclamped position runs the
+     * panel off-screen entirely below ~866px (`screen.width / 2 + 221 +
+     * this.width > screen.width`, solved for `screen.width`). Below that
+     * width the panel overlaps the quickbar instead of vanishing, which is
+     * the same trade-off a real user can still click through.
+     */
+    protected override setPosition(): void {
+        const x = Math.min(G.app.screen.width / 2 + 442 / 2, G.app.screen.width - this.width)
```
wormeyman: **The clamp has no lower bound, so it reintroduces the same failure at the other edge.**

`Math.min(width / 2 + 221, width - this.width)` goes negative once `G.app.screen.width < this.width` (~212px), putting the panel partly off the *left* edge - which is the failure this clamp was added to prevent, just mirrored. `Math.max(0, ...)` around it closes that.

Also, the doc above says that below ~866px the panel "overlaps the quickbar instead of vanishing, which is the same trade-off a real user can still click through". `UIContainer.ts:40` adds `toolsPanel` *after* `quickbarPanel`, so ToolsPanel is on top - the overlap makes the quickbar's rightmost slots **unclickable**, not click-through. Measured at 800x720 the overlap is 33px. The trade-off may still be the right one, but it isn't the one written down.

wormeyman (commented):
Reviewed at `f90e865`. Fourteen inline findings below, plus one convention note.

**Two of them I'd treat as blocking**, and both surface on first use of the feature - consistent with the test plan's manual-verification box still being unchecked:

1. **`ExportDialog`'s pre-selection never fires** (`ExportDialog.ts:43`). `select()` runs in a microtask while the textarea is still `display: none`, so `Ctrl+C` falls through the document `copy` listener to `navigator.clipboard.writeText` - the exact API the dialog exists to bypass.
2. **`exportGuardResult` doesn't report a guard, it runs both actions** (`packages/website/src/index.ts:339`) - a real clipboard write and a real PNG download - and `__fbe_test` is assigned unconditionally, so it ships in production builds.

Four more are reachable UX defects: Escape can't close either dialog once the field has focus, Replace/Append destroy the user's hand-edited text before the async import can fail, the Alt highlight is an opaque fill inserted above the hover and active sprites, and the export field is editable and silently stale. The rest are docs that don't match their code, one dead function, and test-side duplication.

**Convention** (repo `CLAUDE.md`, *Git and PR Conventions*): the PR title carries `#221`, so this squashes to `...UI actions- #221 (#242)` - the rule is to keep the issue number out of the subject and put `Closes #N` in the body. The body says "Re-submission of #221" but has no `Closes #221`, so merging won't close the issue. Separately, the title's "show details" doesn't name anything in the diff - that slot is labelled ALT and toggles entity info overlays.

Nice work on the comment density throughout - the reasoning left at `ROWS`, `maxLength` and `WireSlot` is the right shape for this repo. Several findings below are that the prose and the code drifted apart, not that the prose was unwelcome.

packages/editor/src/UI/controls/functions.ts
```diff
@@ -367,13 +373,51 @@ function rgbToColorSource(r: number, g: number, b: number): ColorSource {
...
+    const size = typeof data.size === 'number' ? data.size : undefined
+    const width = data.width ?? size
+    const height = data.height ?? size
+    if (width === undefined || height === undefined) {
+        throw new Error(`No width/height/size for ${filename}`)
+    }
+
+    const texture = G.getTexture(filename, data.x ?? 0, data.y ?? 0, width, height)
+    const sprite = new Sprite(texture)
+    sprite.scale.set(maxSize / width)
```
wormeyman: **Scaling by width alone stretches any non-square utility sprite, and a declared width of `0` gives `Infinity`.**

`height` is computed on line 399 and then only used for the `undefined` guard and the texture region - it never reaches the scale. All three current call sites happen to be square (`import_slot`/`export_slot` 64x64, `downloading` size 32), so nothing shows today, but this is exported on `F` as a general helper and `utilitySprites` has plenty of non-square entries.

Separately, `data.width ?? size` keeps a literal `0`, which passes the `width === undefined` guard on line 400 and yields `scale = Infinity`. The guard wants to reject non-positive values, not just `undefined`.

`sprite.scale.set(maxSize / Math.max(width, height))` would preserve the aspect ratio and honour the `maxSize` name.

packages/editor/src/Editor.ts
```diff
@@ -21,17 +21,32 @@ import { Dialog } from './UI/controls/Dialog'
...
+/**
+ * A single object rather than trailing positional parameters, so a caller
+ * can supply just `logger` without also passing `quickActions` positionally
+ * ahead of it (or the reverse) - two optional-ish things read far more
+ * clearly as named fields than as an ordered pair, and a third one added
+ * later cannot reshuffle anyone's call site the way a new positional
+ * parameter would.
+ */
+export interface EditorInitOptions {
+    quickActions: QuickActions
```
wormeyman: **`quickActions` is required with no runtime fallback, unlike `logger` right below it.**

`logger` is optional and has a built-in default; this isn't, and nothing guards `G.quickActions` at any read site. A partial options object - or any JS-side caller - leaves it undefined, and the export-image `ActionSlot` throws `Cannot read properties of undefined (reading 'exportImage')` on click, with no message naming the real cause.

`@fbe/editor` is workspace-internal with one consumer today, so this is about failure shape rather than a broken published API. Making it optional and hiding or disabling the four dependent slots when it's absent degrades instead of crashing, and would also let a read-only embed skip the clipboard/FileSaver plumbing entirely.

packages/website/src/index.ts
```diff
@@ -262,6 +323,26 @@ document.addEventListener('paste', (e: ClipboardEvent) => {
...
+    openImportDialog: () => editor.openImportDialog(),
+    /**
+     * `exportString`/`exportImage`'s own guard result - false, with no
+     * clipboard write or file save attempted, whenever the loaded blueprint
+     * is empty. The only part of the two that is safe to call from a spec:
+     * a non-empty blueprint would reach `navigator.clipboard`/`FileSaver`,
+     * neither of which a headless run can assert on. See
+     * tests/quick-actions.spec.ts.
+     */
+    exportGuardResult: () => ({ exportString: exportString(), exportImage: exportImage() }),
```
wormeyman: **This doesn't report the guard - it runs both actions, and it ships in production.**

The doc in `tests/helpers/fbe-test-api.ts` calls it "the only part of either that is safe to call from a spec, since a non-empty blueprint would reach the OS clipboard or a file save". But the body *is* `exportString()` and `exportImage()`. Called with a non-empty blueprint it writes the blueprint string to the OS clipboard and triggers a PNG download.

Two ways that gets reached: a future spec that loads a blueprint first and doesn't read the caveat, and any script on the deployed site - `__fbe_test` is assigned unconditionally at `index.ts:255`, so this is on `window` at fbe.factorygamefan.com.

A hook whose safety depends on the caller knowing the app is in one particular state is the wrong shape. Return the guard itself:

```ts
exportGuardResult: () => ({ exportString: !bp.isEmpty(), exportImage: !bp.isEmpty() }),
```

That's the condition both actions branch on, it pins the same thing the spec asserts, and it can't do anything when it's wrong.

tests/tools-panel.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+/*
+    The corpus's largest file rather than a hardcoded name - what matters is
+    that it's bigger than the 2**20 (1 MiB) cap ImportDialog's textarea used
+    to carry, and picking the largest measured one directly means this stays
+    true even if the corpus changes. Read once at module scope, not per test:
+    it's the input to one test only, but the file is ~2.4 MB and there is no
+    reason to pay that read twice.
+*/
+const LARGEST_BLUEPRINT_FILE = discoverBlueprintFiles().reduce((largest, f) =>
```
wormeyman: **Double-stats every file, and throws at module scope on an empty corpus.**

`fs.statSync(largest.filePath)` re-stats the running maximum on every iteration - 2n syscalls for n files instead of n. Mapping to `{ file, size }` once and reducing over that fixes it.

More importantly, `reduce` with no initial value throws `Reduce of empty array with no initial value` if `discoverBlueprintFiles()` ever returns `[]` - a partial checkout with no `test-blueprints/`, say. At module scope that fails the whole spec file with a message naming neither the corpus nor the spec.

tests/tools-panel.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+}) => {
+    // Sanity on the fixture itself: the whole point is a string past the old
+    // 2**20 (1 MiB) cap, so a corpus that shrank under that would silently
+    // stop testing anything.
+    expect(Buffer.byteLength(LARGEST_BLUEPRINT_STRING)).toBeGreaterThan(2 ** 20)
+
+    await page.evaluate(() => window.__fbe_test.openImportDialog())
+
+    const maxLength = await page.evaluate(() => {
+        const el = [...document.querySelectorAll('textarea')].find(t => t.style.cssText !== '')
```
wormeyman: **`cssText !== ''` excludes nothing, so this can't disambiguate what it's meant to.**

`TextInput`'s constructor sets `position`/`background`/`border`/`outline`/`transformOrigin`/`lineHeight` inline on every instance (`TextInput.ts:99-107`), so `cssText` is non-empty for all of them - the filter is a no-op.

That matters because `ImportDialog` and `ExportDialog` are independent toggles on the same panel and nothing prevents both being open. With two textareas in `document.body`, `.find()` silently picks whichever was appended first and the spec asserts against the wrong field.

This same four-line lookup is repeated at lines 43, 59, 67 and 86, while `quick-actions.spec.ts:67` does the same job with `page.locator('textarea')`. One helper that names the dialog it means would cover both problems.

tests/tools-panel.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+        ) as HTMLTextAreaElement
+        el.value = text
+        el.dispatchEvent(new Event('input', { bubbles: true }))
+    }, LARGEST_BLUEPRINT_STRING)
+
+    const dialog = await page.evaluate(() => window.__fbe_test.topDialogBounds())
+    // ImportDialog.ts: PADDING=12, FIELD_Y=40, FIELD_HEIGHT=90, ROW_Y=142,
+    // ROW_HEIGHT=38, GROUP_GAP=14, REPLACE_Y=194. DescribedButton.ts:
+    // ROW_BUTTON_WIDTH/HEIGHT centre at (+40, +14) from the row's own (x, y).
+    await page.mouse.click(dialog.x + 12 + 40, dialog.y + 194 + 14)
```
wormeyman: **This is a third hand-written copy of `ImportDialog`'s layout, as bare literals.**

`dialog.x + 12 + 40` and `dialog.y + 194 + 14` are `PADDING`, `ROW_BUTTON_WIDTH / 2`, `REPLACE_Y` and `ROW_BUTTON_HEIGHT / 2` - with the derivation living only in the comment above. `ImportDialog.ts` holds copy one; `quick-actions.spec.ts`'s `PADDING`/`FIELD_Y`/`ROW_Y`/`GROUP_GAP` block holds copy two.

Change `GROUP_GAP` from 14 and this clicks the gap between the Paste and Replace rows - a spec that fails by hitting nothing, which reads as a broken button.

The repo's own rule for this (`CLAUDE.md`, *Asking the real game*): "A bound you write down twice is a bound you have to keep in sync, and the one that loses is the one nothing tests." Exporting the layout constants from `ImportDialog.ts` and importing them in both specs would leave one copy.

wormeyman (commented):
Reviewed `7cd1e36e` against the fourteen findings from my last pass. Ten are fixed. I checked each one against the code rather than against the commit message. Four are untouched with no note either way, and one of the fixes introduced a new problem that I measured.

## The two blocking ones are properly fixed

`select()` now defers to `UPDATE_PRIORITY.UTILITY`, which runs below Application's own LOW-priority render step, so `_updateDOMInput` has shown the textarea by the time it fires.

`exportGuardResult` is now `!bp.isEmpty()`. I checked that against both functions: it is the exact condition `exportString` and `exportImage` each open with, so it reports the guard without being able to perform either action.

Both spec files ran in CI rather than being collected and skipped (shards 3 and 4).

## New: the freshness fix costs 310 ms every 500 ms

I asked for this one, so the cost is on me, but it is worth fixing before merge.

`ExportDialog` re-encodes on a 500 ms interval now, and `encode()` (`bpString.ts:249`) is fully synchronous inside its Promise executor. `serialize()`, `JSON.stringify`, `pako.deflate` and base64 all run on the main thread before `resolve()` is called.

Measured on the corpus's own largest blueprint, `test-blueprints/EARN/pocket-base-space-age-v22.1.2.txt`, a 2.42 MB string holding 29.95 MB of JSON:

```
stringify + deflate + base64: median 310 ms (min 300, max 329, 7 runs)
```

That is a lower bound. It leaves out `Blueprint.serialize()`, which I cannot run outside a browser. And `m_LastEncodeAt` is stamped at the top of `refreshText`, before the work rather than after it, so the interval is start-to-start: about 190 ms of idle between 310 ms stalls. Call it 62% of the main thread and 19 dropped frames per cycle.

The dialog deliberately does not block the canvas, which is why the field could go stale at all. So this makes the editor stutter during the one activity the freshness fix exists to support.

`History` already keeps a `historyIndex` that moves on every commit, undo and redo (`History.ts:198`). It is private today, but exposing it, or any monotonic counter, turns the tick into an integer comparison that re-encodes only when something actually changed. That costs nothing while the user is reading rather than editing.

## Four findings unaddressed, with nothing said either way

| # | Site | State |
| --- | --- | --- |
| 6 | `ToolsPanel.ts:151` | The doc still promises "a blank square the same size as a real icon". The code still returns a childless `new Container()`, which draws nothing. |
| 8 | `ToolsPanel.ts:314` | `Math.min` still has no lower bound, so it goes negative below about 212px, which is the failure it was added to prevent, mirrored. The doc still calls the quickbar overlap "click through" when `UIContainer` adds ToolsPanel after the quickbar. |
| 9 | `controls/functions.ts:406` | File untouched. Still `scale.set(maxSize / width)`, so height never reaches the scale, and `data.width ?? size` still passes a literal `0` through the `=== undefined` guard to `Infinity`. |
| 7 | `ToolsPanel.ts:262` | Half done. `placeCells` is extracted and called now, but `generateSlots` still has no call site, and the comment's factual error survives word for word: `Container.destroy()` with no options does not destroy children either. |

I am not asking for all four before merge. 6 and 7 are comment corrections. 8 and 9 are each a one-liner, and 9 sits in a helper exported on `F` for general use.

## Smaller

`noopQuickActions.readClipboardText` both logs through `G.logger` and rejects, where the other four log and resolve to a falsy value. `ImportDialog`'s Paste button catches the rejection and logs it again, so a missing `quickActions` raises two toasts for that one action and one for each of the others.

Convention: the `- #221` is gone from the title, thank you. The body still has no `Closes #221`, and that is the half GitHub reads to close the issue. The title's "show details" also still names nothing in the diff. That slot is ALT.

tests/tools-panel.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+        below ~212px (the panel's own width, `24 + 38*5 - 2` for its 5
+        columns) `screen.width - this.width` goes negative, and nothing
+        stopped that from reaching `position.set`, pushing the panel off the
+        *left* edge instead of merely overlapping the quickbar the way the
+        800px case above does. 150px is inside that range.
+    */
+    await page.setViewportSize({ width: 150, height: 720 })
+
+    const bounds = await page.evaluate(() => window.__fbe_test.toolsPanelBounds())
+    expect(bounds.x).toBe(0)
```
wormeyman: **This is flaky - it failed in a clean full-suite run here, and 2 of 5 times in isolation.** The test is right about the bug; it just reads the answer too early.

`page.setViewportSize` resolves once the browser has resized. It does not wait for pixi to handle the `resize` event, update `G.app.screen.width`, and re-run `setPosition`. So `toolsPanelBounds()` on the very next line often still reports the **1280** layout.

That is what `Received: 861` is, and the number is worth decoding because it hides the defect:

```
min(1280 / 2 + 221, 1280 - 212) = min(861, 1068) = 861
```

It is the un-resized position. It is not off the left edge, it is not negative, and it looks nothing like the `-62` this test exists to catch - so when this fails, the number on screen points away from the mechanism.

**The sibling at :140 has the identical shape and is worse: 3 failures in 6 runs.** It predates this PR, so this is a copied flake rather than a new one, but neither has ever run in CI (the checksum outage above), so both would land green-by-default and then fail about 40% of the time on shard 4.

Measured fix - poll instead of reading once:

```ts
await page.setViewportSize({ width: 150, height: 720 })

await expect
    .poll(() => page.evaluate(() => window.__fbe_test.toolsPanelBounds().x))
    .toBe(0)
```

and for :140, where the assertion is a pair and so has to be polled as one condition:

```ts
await expect
    .poll(() =>
        page.evaluate(() => {
            const b = window.__fbe_test.toolsPanelBounds()
            return b.x >= 0 && b.x + b.width <= 800
        })
    )
    .toBe(true)
```

**12 of 12 runs pass with that**, against 5 failures in 11 without it.

I also ran the control that matters, because a poll that waits for the right answer could just be waiting out a real bug: with the `Math.max(0, ...)` clamp removed, the polled version still fails - and now reports `Received: -62`, the actual off-screen position. So it is strictly stronger than what is here, not more permissive.

tests/quick-actions.spec.ts
```diff
@@ -0,0 +1,418 @@
...
+    if (!at) throw new Error('entity 1 not found in the loaded blueprint')
+    await page.mouse.move(at.x, at.y)
+    await page.keyboard.down('Control')
+    await page.mouse.down({ button: 'right' })
+    await page.mouse.move(at.x + 1, at.y)
+    await page.mouse.up({ button: 'right' })
+    await page.keyboard.up('Control')
+
+    // One history transaction, debounced to exactly one re-encode - not
+    // fired instantly (this is a wait for the debounce delay to elapse, not
```
wormeyman, line 484: **Measured: this is not true. Deleting the debounce leaves the test passing.**

I replaced the whole timer block in `changeTick` with a direct call:

```ts
// MUTATION: debounce removed - fire immediately on every change
this.refreshText({ select: false })
```

Result: `1 passed (2.7s)`. `expect.poll(...).toBe(2)` succeeds the moment the count reaches 2, so it cannot tell "fired after 500 ms" from "fired on the next frame".

Worth saying what the test *does* catch, because that half is good: restoring the pre-fix every-frame encode fails at line 380 with `Received: 4187`. The change-detection is properly pinned. It is only the collapse-a-burst half that is not.

That matters because the burst collapse is the sole justification for `REENCODE_DEBOUNCE_MS`. Delete the constant and the timer and everything here still passes, while the 310 ms encode fires once per transaction during a drag - the cost this commit exists to remove, moved from idle to active rather than eliminated.

A case that would catch it needs two transactions inside one window: delete entity 1, wait well under 500 ms, delete entity 2, then assert the count settles at exactly **2** rather than 3. That is also the only assertion that would notice if the debounce were later changed to leading-edge, which is a change someone may well want - a single edit currently waits a flat 500 ms where the old interval averaged 250 ms.

tests/quick-actions.spec.ts
```diff
@@ -0,0 +1,418 @@
...
+        Nothing about the field's own text can prove the idle half of this:
+        a re-encode of unchanged content is textually identical to no
+        re-encode at all, which is exactly why `encodeCount` exists - see
+        its own doc comment.
+    */
+    await loadBlueprint(page, TWO_CHESTS)
+    await page.evaluate(() => window.__fbe_test.openExportDialog())
+
+    // The initial open-time encode.
+    await expect.poll(() => page.evaluate(() => window.__fbe_test.exportEncodeCount())).toBe(1)
```
wormeyman: **This polls for a value that is already there, and it makes a real failure take 60 seconds to report.**

`m_EncodeCount += 1` is the first line of `refreshText`, which the `ExportDialog` constructor calls synchronously. The constructor runs inside `openExportDialog()`'s own `page.evaluate`, so the count is already 1 when the previous line returns. There is nothing to wait for.

The cost is not style. Running the every-frame mutation, this line took the full 60s default poll timeout to report `Received: 4187` - once the count is climbing it never equals 1 again, so the poll just runs out. Line 385 already does it the right way:

```ts
expect(await page.evaluate(() => window.__fbe_test.exportEncodeCount())).toBe(1)
```

Same change here. (Note this is the opposite call from the ToolsPanel comment above - there the value genuinely arrives late, here it is already present. The question is always whether anything asynchronous stands between the write and the read.)

packages/editor/src/core/History.ts
```diff
@@ -199,6 +199,23 @@ export class History {
     private activeTransaction: Transaction | undefined
     private transactionHistory: Transaction[] = []
 
+    /**
+     * Changes by exactly ±1 on every `commitTransaction`, `undo` and `redo` -
```
wormeyman: **"exactly ±1" is wrong, and I measured how wrong.** None of it breaks `ExportDialog`, but this is new public surface on a core class and the doc is what the next caller reads.

I ran a throwaway vitest against the real class (deleted afterwards):

- **+1 on commit, -1 on undo, +1 on redo.** As documented.
- **Does not move on an empty top-level `commitTransaction`** - the `openTransaction.empty()` early return sits above the increment. Same path as #255.
- **Does not move on the inner commit of a nested transaction.** Correct by design, since the outer commit is the one that counts - but "on every `commitTransaction`" reads otherwise, and the nested case is the common one because `updateValue` opens its own transaction.
- **Jumps by -198 when the length trim fires.** Measured, once, at the 1002nd commit: `[before, after, delta] = [1001, 803, -198]`, from

```ts
if (this.historyIndex > this.MAX_HISTORY_LENGTH) {
    this.transactionHistory.splice(0, this.MAX_HISTORY_LENGTH - this.MIN_HISTORY_LENGTH)
    this.historyIndex = this.transactionHistory.length
}
this.historyIndex += 1
```

`ExportDialog` is fine throughout: it only tests `!==`, and a commit that does not move the index is a commit that changed nothing, so not re-encoding is the right answer.

The risk is a future caller reading "±1" as "a monotonic edit counter" and using `revision - lastSeen` as a number of edits. Something like this says what is actually true:

> Moves whenever a top-level transaction is actually committed, or an undo or redo moves the pointer. Not an edit count and not a state identity - it can jump by more than one when the history-length trim fires, and it does not move for a commit that turned out to be empty.

The paragraph about the undo-then-different-commit collision is good and should stay.

packages/editor/src/UI/ExportDialog.ts
```diff
@@ -0,0 +1,168 @@
...
+ * silently placed on the clipboard. Pre-filled and pre-selected on open, so
+ * a plain Ctrl/Cmd+C is enough once the dialog is up.
+ */
+export class ExportDialog extends Dialog {
+    private readonly m_TextInput: TextInput
+    private m_LastSeenRevision = -1
+    private m_DebounceTimer: ReturnType<typeof setTimeout> | undefined
+    private m_EncodeCount = 0
+
+    /** How many times `refreshText` has actually run a `serialize` +
```
wormeyman: **This counts `refreshText` calls, not encodes.** The increment is the first line of `refreshText`, before `encodeCurrent()` is reached:

```ts
private refreshText({ select }: { select: boolean }): void {
    this.m_EncodeCount += 1
    G.quickActions.encodeCurrent()
```

and on an empty blueprint `encodeCurrent` returns before any serialize happens (`packages/website/src/index.ts:205`):

```ts
function encodeCurrent(): Promise<string | undefined> {
    if (bp.isEmpty()) return Promise.resolve(undefined)
```

So the counter moves for a call that ran neither `serialize` nor `deflate`. Harmless for the one spec reading it today, which uses a loaded blueprint - but the doc is the contract, and the obvious next use of this hook is "does deleting the last entity re-encode", which is exactly where the two diverge.

Either move the increment into the `.then`, where an encode has actually happened, or say what it counts: "how many times `refreshText` has been called".

wormeyman (commented):
Reviewed `c8f505ba` and `98f09316` together. **Nine of the fourteen findings from my earlier passes are now fixed and I checked each against the code rather than the commit message.** The performance fix is right, and all four of the older ToolsPanel/functions.ts findings are genuinely closed.

I verified everything locally, because **CI has never run either commit** - see the first section.

Two things need fixing before merge, and one of them I found by running the suite rather than reading it.

## CI is red repo-wide, and it is not this PR's fault

All four Playwright shards and `checks` fail in 14-29s at `Set up Vite+`:

```
sha256sum: WARNING: 1 computed checksum did NOT match
```

vite-plus 0.3.0 shipped on 2026-08-24 and rewrote `install.sh`, so the pinned hash stopped matching under an unchanged `VP_VERSION`. That is the pin working. **#260 fixes it** and is already green on `checks`. Until it lands, a red `Set up Vite+` here says nothing about this PR.

## Verified fixed

`c8f505ba` - `ExportDialog` polls `History.revision` (an integer compare) and debounces a re-encode 500 ms after the blueprint stops changing. The idle 310 ms stall every 500 ms is gone. I also went looking for the staleness hole this approach could open and it is not reachable: `loadBlueprint` calls `Dialog.closeAll()` before anything can observe a new blueprint's reset `historyIndex`, and `history.reset()` runs only in the `Blueprint` constructor.

`98f09316` - all four older findings, checked individually:

| # | Fix | Verified |
| --- | --- | --- |
| 6 | `safeIcon` doc now says "empty, childless `Container`" | Matches the code |
| 7 | `generateSlots` doc no longer claims `destroy()` cascades | Correct, and `placeCells` really does `child.destroy()` on each removed child, so the new claim holds too |
| 8 | `Math.max(0, Math.min(...))` | Mutation-checked - removing it gives `x = -62` at 150px |
| 9 | `!width \|\| !height`, and `maxSize / Math.max(width, height)` | Catches the literal `0`, and fits both axes |

Local gates on `98f09316`: `vp check` clean (0 lint/type errors), `vp test` 180/180, full Playwright **212 passed, 1 failed** - and that one failure is the subject of the next section.

## Blocking: the new ToolsPanel test is flaky, and so is the one it was modelled on

`tools-panel.spec.ts:155` failed in a clean full-suite run on unmutated code, with `Received: 861`.

Measured in isolation:

| Test | Result |
| --- | --- |
| `:155` "does not run off the left edge" (new in `98f09316`) | **2 failures in 5 runs** |
| `:140` "stays on screen at a narrow viewport width" (pre-existing) | **3 failures in 6 runs** |

So this is a pre-existing flake that the new test copied, not one it introduced - but neither has ever run in CI, so both would land and then fail on shard 4 roughly 40% of the time, in a spec nobody touched.

The mechanism, and the fix I measured, are inline.

## Blocking: the ExportDialog test does not guard the debounce, and says it does

Deleting the debounce entirely leaves the test **passing, in 2.1s**. Inline.

## Two doc comments overstate their code

`History.revision` and `ExportDialog.encodeCount`. Comment-only, both measured, both inline. `revision` is new public surface on a core class, which is why it is worth getting exactly right.

## Still open from earlier passes

`noopQuickActions.readClipboardText` rejects where its four siblings resolve to a falsy value, so a missing `quickActions` raises two toasts for that one action and one for each of the others. And the body still has no `Closes #221`, which is the half GitHub reads to close the issue.

wormeyman:
@claude review

wormeyman:
Re-reviewed at `9338da48`. Almost everything from the previous three rounds is fixed with the mechanism removed rather than the wording satisfied, which is the harder and better way to do it. The newest commit is a genuine deterministic fix, not a settle-wait: `x >= 0` alone was already true at the stale 1280px layout, so the old poll resolved on its first check and added no wait at all. The joint condition can actually fail. That is the right shape.

Three things left. One is user-facing.

## Closing either dialog leaves Ctrl+C and Ctrl+V silently dead

Both handlers open with `if (document.activeElement !== CANVAS) return` (`packages/website/src/index.ts:305` and `:311`). When a dialog is destroyed, `TextInput._onRemoved` removes the focused `<input>` from `document.body`, which resets `document.activeElement` to `<body>`. Nothing in `packages/` ever focuses the canvas back. The only `.focus()` calls in the whole tree are `TextInput.ts:195` and `:203`.

So: open Export, which focuses the textarea by design, press Escape, then press Ctrl+C. Nothing happens. No toast, no error. The user gets it back only by clicking the canvas.

The canvas already has `tabindex="1"`, so a `CANVAS.focus()` on dialog close is the whole fix. `quick-actions.spec.ts:242` presses Escape and then asserts only `openDialogCount() === 0`, so no test can see this today.

## The empty-field test cannot fail for its stated reason

`quick-actions.spec.ts:185` is written to cover `runImport`'s up-front empty guard. Delete that guard and every assertion still passes. `importReplace('')` falls through to `new URL('https://')` and throws, `importReplace`'s own `.catch` swallows it, so there is no `pageerror`; the dialog stays open; the blueprint is untouched. The only observable difference is the toast text. Asserting that would close it.

## One comment is wrong in a way that invites a real bug

`ToolsPanel.ts:319` says the quickbar overlap below ~866px is "the same trade-off a real user can still click through". `UIContainer.ts:40` adds the panel after the quickbar, so it draws on top, and pixi hit-tests children in reverse. Those slots are dead, not click-through. Someone reading that comment and widening the panel or relaxing the clamp would make it worse, and nothing in `tests/` hit-tests over the quickbar, so it would land green.

Take or leave: the panel is a noticeably larger permanent dead zone over the canvas than `WiresPanel` was, and it is added after `dialogsContainer`, so at short viewport heights it covers the bottom of a centred dialog.

wormeyman (commented):
Details in the comment above. Short version: almost everything from the previous rounds is fixed with the mechanism removed rather than the wording satisfied, and the newest commit is a real deterministic fix rather than a settle-wait.

Three things left. The one worth doing before merge is the focus bug: closing either dialog leaves Ctrl+C and Ctrl+V silently dead, because nothing focuses the canvas back. `CANVAS.focus()` on close is the whole fix.

wormeyman (commented):
I drove this one rather than only reading it: a worktree at the PR head, both dev servers up, and mutations applied one at a time.

The mechanism is right. The canvas carries `tabindex="1"` (`packages/website/index.html:90`) and `G.app.canvas` is that same element, so it is the one the clipboard guards compare against. `close()` filters `s_openDialogs` on its first line, so `anyOpen()` after `destroy()` is not off by one. `destroy()` removes the `<input>` synchronously, so nothing can undo the `focus()` afterwards. `#editor` has `outline: none`, so no focus ring appears.

The two tests carry their weight. I measured the mutation table rather than reasoning about it:

| mutation | test 1 (last dialog) | test 2 (one still open) |
|---|---|---|
| remove `G.app.canvas.focus()` | fails | passes |
| drop the `!anyOpen()` guard | passes | fails |

Each kills exactly one. The end-to-end claim holds too: after opening and closing ExportDialog with Escape, a real Ctrl+C raises the "copied to clipboard" toast with no canvas click first.

## One regression

With a dialog open, the BP Book Index arrow keys stop working after one press.

`settingsPane.ts:47` hangs its own keydown handler on that input. ArrowUp calls `changeBookIndex` -> `editor.loadBlueprint` -> `Dialog.closeAll()` (`Editor.ts:415`) -> the last close -> `G.app.canvas.focus()`. The live `<input>` loses focus, so the second arrow reaches the editor's keybinds instead of the box.

| | focus after 1st arrow | index after 1st / 2nd |
|---|---|---|
| this PR | `CANVAS` | 1 / **1** |
| commit reverted | `INPUT` | 1 / **2** |
| no dialog open (control) | `INPUT` | 1 / 2 |

Nothing in the suite can see it. `git grep ArrowUp tests/` is empty.

## Fix

Only claim focus that is genuinely orphaned:

```diff
-        if (!Dialog.anyOpen()) G.app.canvas.focus()
+        if (!Dialog.anyOpen() && document.activeElement === document.body) {
+            G.app.canvas.focus()
+        }
```

Reading `document.body` separates "the field that had the focus was just destroyed" from "something else still holds it". The first is the case the hand-back is for, and the only one it should touch. The comment above the guard needs the second half explained; I can hand over my wording if that helps.

With this applied, `vp check`, `vp test` (188) and the full Playwright suite (218) all pass. Reverting just the `activeElement` half fails the second test in the spec below and nothing else.

<details>
<summary>tests/settings-pane-book-index.spec.ts (new file)</summary>

```ts
import { test, expect } from '@playwright/test'
import {
    encodeBlueprintBook as encodeBook,
    packVersion as version,
} from './helpers/encode-blueprint'
import { loadBlueprint, waitForEditor } from './helpers/fbe-test-api'

/*
    The settings pane's BP Book Index box, and the first coverage its arrow
    keys have ever had - `git grep ArrowUp tests/` was empty before this file.

    `packages/website/src/settingsPane.ts` hangs its own `keydown` listener on
    that dat.gui `<input>` so ArrowUp/ArrowDown step the index. Stepping is
    the whole reason the listener exists, and stepping means the *second*
    press has to work as well as the first - which is what makes this
    reachable at all, and what nothing else here can see.

    Written for a regression the #242 focus hand-back introduced (see
    `Dialog.close()`). Each step calls `changeBookIndex` ->
    `editor.loadBlueprint` -> `Dialog.closeAll()`, so with any dialog open the
    first arrow closes it, and an unconditional `G.app.canvas.focus()` on that
    close pulled the focus off the live <input>. The second arrow then reached
    the editor's keybinds instead of the box. Measured 0 -> 1 -> 1 against the
    0 -> 1 -> 2 the control gives.

    Two things this spec is shaped by.

    The pair is a real pair, not a test and a restatement. Only the second
    test fails against the unconditional guard; the first passes there, and
    both pass with the focus hand-back deleted outright. So the control says
    the arrows work at all and the treatment says a dialog does not break
    them, and neither one alone distinguishes the three states.

    It does not call `suppressOverlays`. That helper sets `pointer-events:
    none` on `.dg.main`, which is this pane - the very thing the test has to
    click. The pane is bottom *left* and the toasts are bottom *right*, so
    there is nothing here for the suppression to protect against anyway.

    The book entries carry a different entity count each, and that is the
    synchronisation. `setValue` writes the box synchronously and only then
    starts an async load, so waiting on the box's own value proves nothing
    about whether the load - and the `closeAll` inside it - has run yet.
    Pressing the second arrow into that gap passes whatever the guard does.
    `entityContainerCount()` moves only when a load finishes, and it names
    *which* entry finished.
*/

const VERSION = version(2, 0, 55)

const chests = (n: number): Record<string, unknown>[] =>
    Array.from({ length: n }, (_, i) => ({
        entity_number: i + 1,
        name: 'wooden-chest',
        position: { x: i + 0.5, y: 0.5 },
    }))

/** Entry i holds i+1 chests, so entityContainerCount() names the loaded entry. */
const BOOK = encodeBook({
    item: 'blueprint_book',
    version: VERSION,
    active_index: 0,
    blueprints: [0, 1, 2].map(index => ({
        index,
        blueprint: { item: 'blueprint', version: VERSION, entities: chests(index + 1) },
    })),
})

type Page = import('@playwright/test').Page

const bpIndexInput = (page: Page) =>
    page.locator('.dg .property-name:has-text("BP Book Index")').locator('..').locator('input')

const activeTag = (page: Page): Promise<string | undefined> =>
    page.evaluate(() => document.activeElement?.tagName)

/** Waits for the load the arrow started to finish, by the entry it landed on. */
const waitForEntry = (page: Page, index: number): Promise<unknown> =>
    page.waitForFunction(n => window.__fbe_test.entityContainerCount() === n, index + 1)

test.beforeEach(async ({ page }) => {
    await waitForEditor(page)
})

test('CONTROL: the arrow keys step the book index twice over, with no dialog involved', async ({
    page,
}) => {
    await loadBlueprint(page, BOOK)
    const input = bpIndexInput(page)
    await input.click()
    expect(await activeTag(page)).toBe('INPUT')

    await page.keyboard.press('ArrowUp')
    await waitForEntry(page, 1)
    expect(await activeTag(page)).toBe('INPUT')

    await page.keyboard.press('ArrowUp')
    await waitForEntry(page, 2)

    await expect(input).toHaveValue('2')
})

test('a dialog closing on the first arrow does not steal the focus off the index box', async ({
    page,
}) => {
    await loadBlueprint(page, BOOK)
    await page.evaluate(() => window.__fbe_test.openExportDialog())
    expect(await page.evaluate(() => window.__fbe_test.openDialogCount())).toBe(1)

    const input = bpIndexInput(page)
    await input.click()
    expect(await activeTag(page)).toBe('INPUT')

    await page.keyboard.press('ArrowUp')
    await waitForEntry(page, 1)

    // Load-bearing: without this the test could pass on a dialog that never
    // closed, which is the one state that cannot exercise the guard at all.
    expect(await page.evaluate(() => window.__fbe_test.openDialogCount())).toBe(0)
    expect(await activeTag(page)).toBe('INPUT')

    await page.keyboard.press('ArrowUp')
    await waitForEntry(page, 2)

    await expect(input).toHaveValue('2')
})
```

</details>

## The guard's stated reason is not true

The comment says that while another dialog is open, "its field is what the keyboard should be talking to". It is not. Only ExportDialog focuses its own field (`ExportDialog.ts:168`); ImportDialog never does.

I drove the exact case the second test sets up, open Import, open Export, Escape, and `document.activeElement.tagName` is `BODY`, with the Import textarea on screen and unfocused. Typing reached no field at all. The keys went to the editor's keybinds, and the `E` in my test string opened the inventory dialog.

That state has the same dead keyboard the commit removes elsewhere. It predates the commit, so it is not a regression. But the second test pins it as intended behaviour, and the comment justifies it with something the code does not do. Either correct the comment or focus the surviving field.

## Minor

- The commit message says TextInput's two calls are "the only `.focus()` in the tree". Two more live in `tests/`. The in-code comment says "in the package" and is correct.
- The mutation table is in the commit message. `paste-placement.spec.ts` and `splitter-wires.spec.ts` both keep theirs in the file header, which is what someone reads before deleting a test that looks like it always passes.
- `page.evaluate(() => window.__fbe_test.openExportDialog())` now appears 7 times raw, while `openImportDialog` has a helper.

One more that I could not verify: `s_openDialogs` leaks if a dialog subclass constructor throws, which would leave `anyOpen()` permanently true and disable the hand-back for the rest of the session. I found no reachable path to it.

PR #245, opened 2026-08-19
Drop the apt install for chromium's libraries; assert them instead

wormeyman:
`npx playwright install-deps chromium` was the least reliable step in this workflow, and on a cache hit it had nothing to do. This deletes it and asserts the thing it was there to guarantee.

Both halves are measured. Neither was assumed, and the first attempt at this PR got it wrong - see the history at the bottom.

## It was unreliable

44 runs of the step across 12 CI runs, from the API rather than from any single run:

| | `install-deps` | `Run Playwright specs` |
|---|---|---|
| min | 11s | 96s |
| median | **17s** | - |
| max completed | **686s** | 226s |
| over 5 minutes | 5 of 42 | 0 of 39 |
| never finished | **2** | 0 |

Bimodal rather than noisy: 32 of 42 under 40s, then a long tail with nothing in between, while the specs either side stay tight. The two that never finished took the 20-minute job timeout with them, and with it `deploy`, which waits on `e2e`. That is what stopped #244 reaching fbe.factorygamefan.com twice, on a commit whose every other job passed.

## It was unnecessary

With the step gone, all four shards passed and `ldd` reports no missing shared libraries. The `ubuntu-24.04` image already carries what the cached chromium needs.

The comment being deleted is not wrong, it just stops one question early. The OS packages genuinely do not survive the cache - only the browser does. Nobody had checked whether they needed to.

## Retrying it was the wrong fix, and the log says why

The first version of this PR wrapped the step in `timeout` and retried. It made CI worse: all four shards failed at ~4m40s where most had passed in 17s.

```
16:19:35  Get:1 file:/etc/apt/apt-mirrors.txt Mirrorlist    <- apt stalls here
16:23:34  ##[warning]install-deps attempt 1 did not finish within 240s; retrying
16:23:47  E: Could not get lock /var/lib/apt/lists/lock. It is held by process 2428 (apt-get)
```

`timeout` kills its own child, `npx`. The root-owned `apt-get` underneath survives and keeps the lock, so attempts 2 and 3 die instantly. Identical on all four shards. **A timeout around a `sudo` call does not bound the `sudo` call.**

Worth noting all four hung on attempt 1 in that run, against a median of 17s, which points at the mirror having a bad period rather than at per-runner variance - so no retry count would reliably have helped either.

## What replaces it

An assertion, not an install: `ldd` over the cached chromium binaries. About a second, and it cannot hang. If a runner image ever drops a library, this fails here naming the binary, instead of surfacing as chromium failing to launch inside an unrelated spec.

**The expected binary count is part of the assertion, and is the half worth keeping.** The first draft of the check searched for `headless_shell` when the file is `chrome-headless-shell`, so it probed one binary instead of two and reported a clean result for a binary the specs do not run - `headless: true` in playwright.config.ts means the headless shell is what they use. A missing-library count means nothing without the number of things it looked at.

Tested under `bash -eo pipefail`, which is what Actions uses:

| Case | Expected | Result |
|---|---|---|
| both binaries, no missing libs | exit 0 | pass |
| both binaries, a library missing | **exit 1** | pass |
| only `chrome` present, count short | **exit 1** | pass |
| headless shell under the old wrong name | **exit 1** | pass |
| no binaries found at all | **exit 1** | pass |

The fourth case reproduces the exact mistake above and now fails instead of passing quietly.

`timeout-minutes` is unchanged at 20 - nothing slow is left in the job. The net diff is one step replaced and nothing else.

## Note on how this PR got here

Three commits: a retry that made CI worse, an experiment that measured why and whether the step was needed, and this. Kept rather than squashed away in the branch, because the useful part is the sequence - a plausible fix, a measurement that refuted it, and a smaller change that follows from what was measured. It will squash to one commit on merge.

The transferable bit: the test I wrote before pushing the retry stubbed the command out, so it verified the loop's control flow and was blind to apt's global system state, which is the thing that actually breaks. A stub tests the code around the call, never the call's interaction with the system.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,14 +246,7 @@ jobs:
     # backstop against a hung suite rather than an estimate: a shard that has
     # genuinely stopped making progress should be killed, and one that is merely
     # slow should not.
-    #
-    # 20 -> 25 because the retry around `Install browser system dependencies`
-    # below bounds that step at 12 minutes rather than the 20 it once ran to,
-    # and 12 + ~1 setup + ~4 specs does not fit under 20. This is not a way of
-    # tolerating a slower suite: the specs are measured at 96-226s and nothing
-    # here expects that to move. If a shard reaches 25 minutes, something is
-    # hung, which is the same thing the 20 meant.
-    timeout-minutes: 25
+    timeout-minutes: 20
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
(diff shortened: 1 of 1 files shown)


wormeyman:
## Course correction: this PR is now an experiment

The retry I pushed first **made CI worse** - all four shards failed at ~4m40s where most previously passed in 17s. The run that broke it also explained why, and the explanation changes the question the PR should be asking.

### What actually happens

```
16:19:35  Get:1 file:/etc/apt/apt-mirrors.txt Mirrorlist    <- apt hangs here
16:23:34  ##[warning]install-deps attempt 1 did not finish within 240s; retrying
16:23:47  E: Could not get lock /var/lib/apt/lists/lock. It is held by process 2428 (apt-get)
```

`timeout` kills its own child, `npx`. The root-owned `apt-get` underneath survives and keeps the lock, so attempts 2 and 3 fail instantly. `dpkg --configure -a` cannot help - the lock is held by a live process, not a broken transaction. Identical on all four shards (PIDs 2428, 2434, 2553, and one more).

Two things worth recording. All four shards hung on **attempt 1** in that run, against a historical median of 17s, which points at the mirror having a bad period rather than at per-runner variance. And the test I wrote before pushing stubbed the command, so it verified the control flow and could not see the thing that breaks - apt's global system state. A stub tests the loop, not the interaction.

### The question that replaces it

The step only runs on a cache hit, and the comment above it asserts that the OS packages do not survive the cache - true - without ever checking whether the `ubuntu-24.04` image already carries them. If it does, the fix is to **delete** the step, not retry it, and apt leaves the critical path entirely.

So this push asks the linker rather than apt: `ldd` each cached chromium binary and report which shared libraries are actually missing. A second, no apt, cannot hang. The specs then run regardless, so one run gives two independent readings - what `ldd` reports, and whether the suite passes without the install - and they should agree.

The probe cannot fail the job on purpose; a missing library has to present as the specs failing, which is the reading being taken. Tested under `bash -e` across no-binaries-found, none-missing and libs-missing - all three exit 0.

**Do not merge this commit.** Once the reading is in, this becomes either a deletion of the step or a retry that kills the orphaned apt process, and the PR gets rewritten to match.

PR #258, opened 2026-08-25
Handle blueprints without generated icons

### LINKED ISSUE #254: generateIcons throws on a blueprint whose entities are all non-minable, so it cannot be exported
`Blueprint.generateIcons()` picks the icon for a blueprint that carries none, and it reads the top entry of its sorted list without checking that the list has one:

https://github.com/FactoryGameFan/factorio-blueprint-editor/blob/wormeyman-space-age-support/packages/editor/src/core/Blueprint.ts#L772

`getIconPairs` skips every entity whose `Entity.getItemName` is undefined, which is `FD.entities[name]?.minable?.result`. So a blueprint whose entities are *all* non-minable reduces to an empty list, and `iconPairs[0][0]` throws.

## Reachable by hand

20 of the 155 entities in `data.json` have no `minable.result`, which matches the figure in `CLAUDE.md`. Two of them can be placed straight from the inventory, because an item names them as its `place_result`:

```
captive-biter-spawner  <- item "captive-biter-spawner"
space-platform-hub     <- item "space-platform-hub"
```

`space-platform-hub` is the centre of every space platform, so "a blueprint holding just the hub" is an ordinary thing to make rather than a contrived one.

Replaying the reducer against the committed `data.json`, with `transport-belt` as a control:

```
[{name:"space-platform-hub"}]    -> iconPairs=[]  THROWS -> TypeError: Cannot read properties of undefined (reading '0')
[{name:"captive-biter-spawner"}] -> iconPairs=[]  THROWS -> TypeError: Cannot read properties of undefined (reading '0')
[{name:"transport-belt"}]        -> iconPairs=[["transport-belt",0]]  icons.set(1, "transport-belt")
```

## Repro

Open the editor on an empty blueprint, place a `space-platform-hub` from the inventory, press Ctrl+C.

`serialize()` calls `generateIcons()` whenever `icons.size` is 0, `encode()` catches the throw, and you get a toast that names a TypeError rather than anything about icons. No blueprint string reaches the clipboard.

The tile branch at line 785 has the same shape, though `Tile.getItemName` may cover every tile in `data.json`. Worth checking rather than assuming.

## Why it is worth fixing now

Today the cost is one failed export. That changes if #243 lands as it currently stands, because it routes these writes through `History` and the throw then happens inside an open transaction, which stalls the transaction count and disables undo and redo for the rest of the session. I have asked for a guard on that PR, but the missing guard is here on the base branch and predates it.

See also #255 for the related `History` weakness that turns this from a caught error into a permanent one.

## Suggested fix

Return early when `iconPairs` is empty rather than writing an icon. An entity with no item to show is the case `getIconPairs` already handles one row at a time, so a blueprint made only of those is the same case at the whole-blueprint level, and "no auto icon" is a reasonable answer.

A regression test is cheap and needs no browser for the reducer half. The whole-path half wants a synthetic blueprint, the way `tests/helpers/encode-blueprint.ts` builds one, since nothing in the corpus is made only of non-minable entities.

byalex33:
## Summary

Fix blueprint export when automatic icon generation has no item-backed candidate, while keeping the serialized blueprint valid against the editor schema.

A blueprint containing only non-minable entities, such as `space-platform-hub`, produces an empty item-icon candidate list. `generateIcons()` previously dereferenced the first missing candidate and threw before the blueprint string reached the clipboard.

## Changes

- Guard both entity and tile icon-candidate lists before their indexed reads.
- When an entity blueprint has no item-backed icon, use the first entity's own name as an `entity`-type signal. For `space-platform-hub`, this serializes as `{ index: 1, signal: { type: entity, name: space-platform-hub } }`.
- Keep the tile guard defensive and document why: every tile in current data has an item, and Factorio has no tile signal type for a schema-valid prototype-name fallback.
- Replace the empty-icon assertion with a full export/re-import regression test that confirms the icon survives and no validation warning is produced.
- Document why the test's synthetic `loadData()` call is isolated safely by Vitest's per-file module state.

## Verification

- Focused `Blueprint.test.ts` — 1 test passed
- `vp check --fix` — 179 files, no warnings, lint errors, or type errors
- `vp test` — 16 files, 188 tests passed
- `git diff --check`

Closes #254

### DIFF
--- a/packages/editor/src/core/Blueprint.test.ts
+++ b/packages/editor/src/core/Blueprint.test.ts
@@ -1,0 +1,46 @@
+import { beforeAll, describe, expect, it } from 'vite-plus/test'
+import { Blueprint } from './Blueprint'
+import { loadData } from './factorioData'
+
+beforeAll(() => {
+    loadData(
+        JSON.stringify({
+            items: {},
+            fluids: {},
+            signals: {},
+            recipes: {},
+            entities: {
+                'space-platform-hub': {
+                    type: 'space-platform-hub',
+                    name: 'space-platform-hub',
+                    collision_box: [
+                        [-1, -1],
+                        [1, 1],
+                    ],
+                },
+            },
+            tiles: {},
+            inventoryLayout: [],
+            utilitySprites: {},
+            utilityConstants: {},
+            guiStyle: {},
+            defines: {},
+        })
+    )
+})
+
+describe('Blueprint icon generation', () => {
+    it('serializes without icons when every entity is non-minable', () => {
+        const blueprint = new Blueprint({
+            entities: [
+                {
+                    entity_number: 1,
+                    name: 'space-platform-hub',
+                    position: { x: 0, y: 0 },
+                },
+            ],
+        })
+
+        expect(blueprint.serialize().icons).toEqual([])
+    })
+})
--- a/packages/editor/src/core/Blueprint.test.ts
+++ b/packages/editor/src/core/Blueprint.test.ts
@@ -1,8 +1,11 @@
 import { beforeAll, describe, expect, it } from 'vite-plus/test'
 import { Blueprint } from './Blueprint'
+import { encode, getAndClearLoadWarnings, getBlueprintOrBookFromSource } from './bpString'
 import { loadData } from './factorioData'
 
 beforeAll(() => {
+    // loadData permanently replaces FD's accessors. Vitest's per-file module
+    // isolation keeps this synthetic dataset from leaking into other test files.
     loadData(
         JSON.stringify({
             items: {},
@@ -30,7 +33,7 @@ beforeAll(() => {
 })
 
 describe('Blueprint icon generation', () => {
-    it('serializes without icons when every entity is non-minable', () => {
+    it('round-trips with an entity icon when every entity is non-minable', async () => {
         const blueprint = new Blueprint({
             entities: [
                 {
@@ -41,6 +44,17 @@ describe('Blueprint icon generation', () => {
             ],
         })
 
-        expect(blueprint.serialize().icons).toEqual([])
+        const icons = [
+            {
+                index: 1,
+                signal: { type: 'entity' as const, name: 'space-platform-hub' },
+            },
+        ]
+        expect(blueprint.serialize().icons).toEqual(icons)
+
+        const roundTripped = await getBlueprintOrBookFromSource(await encode(blueprint))
+
+        expect(getAndClearLoadWarnings()).toEqual([])
+        expect(roundTripped.serialize().icons).toEqual(icons)
     })
 })
--- a/packages/editor/src/core/Blueprint.ts
+++ b/packages/editor/src/core/Blueprint.ts
@@ -769,6 +769,7 @@ class Blueprint extends EventEmitter<BlueprintEvents> {
                 (a, b) => getItemScore(b) - getItemScore(a)
             )
 
+            if (iconPairs.length === 0) return
             this.icons.set(1, iconPairs[0][0])
             if (
                 iconPairs[1] &&
@@ -782,6 +783,7 @@ class Blueprint extends EventEmitter<BlueprintEvents> {
                 (a, b) => b[1] - a[1]
             )
 
+            if (iconPairs.length === 0) return
             this.icons.set(1, iconPairs[0][0])
         }
     }
--- a/packages/editor/src/core/Blueprint.ts
+++ b/packages/editor/src/core/Blueprint.ts
@@ -762,7 +762,10 @@ class Blueprint extends EventEmitter<BlueprintEvents> {
                 (a, b) => getItemScore(b) - getItemScore(a)
             )
 
-            if (iconPairs.length === 0) return
+            if (iconPairs.length === 0) {
+                this.icons.set(1, this.entities.valuesArray()[0].name)
+                return
+            }
             this.icons.set(1, iconPairs[0][0])
             if (
                 iconPairs[1] &&
@@ -776,6 +779,8 @@ class Blueprint extends EventEmitter<BlueprintEvents> {
                 (a, b) => b[1] - a[1]
             )
 
+            // Defensive only: every current tile has an item, and Factorio has
+            // no tile signal type to use as a schema-valid fallback.
             if (iconPairs.length === 0) return
             this.icons.set(1, iconPairs[0][0])
         }


packages/editor/src/core/Blueprint.test.ts
```diff
@@ -0,0 +1,46 @@
...
+            entities: [
+                {
+                    entity_number: 1,
+                    name: 'space-platform-hub',
+                    position: { x: 0, y: 0 },
+                },
+            ],
+        })
+
+        expect(blueprint.serialize().icons).toEqual([])
```
wormeyman: **This pins the schema-invalid shape as the expected one.**

`blueprintSchema.json` gives `icons` `"minItems": 1`, and `icons` is in the blueprint definition's `required` list - so `[]` is not a valid value and neither is omitting the key.

Measured, a blueprint that serializes this way logs on re-import:

```
/blueprint/icons  minItems  "must NOT have fewer than 1 items"
Blueprint had validation warnings (loaded anyway)
```

An assertion that would have caught it is the round trip rather than the model: serialize, feed the string back through the loader, and assert no validation warning was raised. As written, this test passes for output the editor itself rejects.

Also worth a note in the file about `loadData`: it permanently swaps every `FD` accessor for a plain data property on first write, so a synthetic payload here relies on vitest isolating module state per file. That is true today and worth writing down, since this is the first unit test in the repo to take that route - everything else FD-dependent goes through Playwright.

packages/editor/src/core/Blueprint.ts
```diff
@@ -769,6 +769,7 @@ class Blueprint extends EventEmitter<BlueprintEvents> {
                 (a, b) => getItemScore(b) - getItemScore(a)
             )
 
+            if (iconPairs.length === 0) return
```
wormeyman: The guard is in exactly the right place - after the sort, before the indexed read. The problem is what it leaves behind: `this.icons` stays empty, and `serialize()` maps that to `icons: []`, which the editor's own schema rejects (`minItems: 1`, and `icons` is `required`).

An entity-type fallback fits what `serialize()` already does. `getItemTypeForBp` has the arm:

```ts
if (FD.entities[name]) return 'entity'
```

and the schema's `signal.type` enum includes `entity`. So rather than returning empty, setting the icon to the entity's own name yields `{index: 1, signal: {type: 'entity', name: 'space-platform-hub'}}` - schema-valid, and closer to the game, which always gives a blueprint an icon.

Same applies to the tile branch below, which has the same guard and the same consequence. Note that branch needs an all-non-minable *tile* to reach, which the current test does not construct - so only one of the two guards is actually exercised.

wormeyman (commented):
Thanks for this - the crash in #254 is real and both guards are in the right place. But the fix trades a throw for an export the editor's own schema rejects, and I'd like that closed before it lands.

## What I measured

I loaded a blueprint holding one `space-platform-hub` (no `minable.result`, so `getIconPairs` skips it), serialized it through the editor, and re-imported the result:

```
SERIALIZED_ICONS=[]

Blueprint validation warnings (loading anyway):
  instancePath: /blueprint/icons
  keyword:      minItems
  message:      "must NOT have fewer than 1 items"
Blueprint had validation warnings (loaded anyway)
```

So it no longer crashes - good - but every round trip of such a blueprint now raises a validation warning toast, and the string we hand the user is one we would reject on the way back in.

`blueprintSchema.json` says:

```json
"icons": { "type": "array", "minItems": 1, "maxItems": 4, ... }
```

and `icons` is also in the blueprint definition's `required` list. **So there is no schema-valid output under this approach** - `[]` fails `minItems`, and omitting the key fails `required`. The early `return` leaves `this.icons` empty and `serialize()` maps that straight to `icons: []`.

`Blueprint.test.ts` currently pins that shape:

```ts
expect(blueprint.serialize().icons).toEqual([])
```

which locks in the invalid output as the expected one.

## A fix that fits what is already there

Fall back to the entity's own name as an **entity-type signal**. `serialize()`'s `getItemTypeForBp` already has the arm for it:

```ts
if (FD.entities[name]) return 'entity'
```

and the schema's `signal.type` enum already includes `entity`. So setting `this.icons.set(1, <entity name>)` when no item-backed icon exists produces `{index: 1, signal: {type: 'entity', name: 'space-platform-hub'}}`, which satisfies both `required` and `minItems`, and matches the game's own behaviour of always giving a blueprint an icon.

Relaxing the schema instead is the other option, but it is the bigger claim - it says the format permits an icon-less blueprint, which is worth checking against the game before asserting.

Either way the test wants to assert the round trip, not just the model: serialize, re-import, and assert **no** validation warning. That is the part that would have caught this.

## Two smaller things

**The unit test is the first here to call `loadData` with a synthetic payload**, which is a nice way to test FD-dependent code without a browser - `tests/` normally needs Playwright for this. Worth a sentence in the file saying why it is safe: `loadData` permanently swaps every `FD` accessor for a plain data property on first write (see `factorioData.ts`), so this relies on vitest isolating module state per file. If that ever changes, this leaks a one-entity `FD` into every other unit test in the same worker.

**Both guards are correct but only one is reachable in the shape you tested.** The second (`else if (!this.tiles.isEmpty())`) needs an all-non-minable *tile*, which is a different fixture. Worth one test each, or a comment saying the tile arm is defensive.

## What is fine

Both `if (iconPairs.length === 0) return` placements are right - after the sort, before the indexed read, in each branch independently. `vp check` clean and `vp test` 185/185 locally.

byalex33:
@wormeyman Addressed your review in `17a340ed`:

- Non-minable entity blueprints now fall back to the entity's own name, which serializes as a schema-valid `entity` signal instead of `icons: []`.
- The regression test now performs the full encode/re-import round trip and asserts that `getAndClearLoadWarnings()` is empty.
- The test documents the `loadData()`/Vitest file-isolation assumption.
- The tile guard is documented as defensive: current tiles all have item representations, and Factorio exposes no tile signal type for an equivalent fallback.

Verification is clean: `vp check --fix` passed across 179 files, and `vp test` passed all 188 tests. Please take another look when you have a chance.

wormeyman (approved):
This is the fallback I suggested, and I checked it against the real `data.json` rather than only the synthetic FD the test builds.

`space-platform-hub` serializes as `{"type":"entity","name":"space-platform-hub"}` and round-trips with no load warnings. `entity` is in the schema's signal type enum, and `itemFluidSignalRecipeEntityName` is a no-op keyword, so nothing rejects the name either. That closes the `icons: []` problem from my last pass.

Three things I checked rather than assumed:

- The `entities.isEmpty()` guard above means `valuesArray()[0]` cannot be undefined.
- The "Defensive only" comment on the tile branch is accurate. 0 of the 22 tiles in `data.json` lack an item, so that `return` really is unreachable. It would have re-introduced the original bug if it were not.
- 20 entities have no `minable.result`, and `space-platform-hub` is one of them, so the test picks a real case rather than an invented one.

`vp check` clean, `vp test` 188/188 locally.

One note, not a blocker and not your code. `captive-biter-spawner` is also one of the 20, and it comes out as `{"type":"recipe",...}` instead of `entity`, because `getItemTypeForBp` checks `FD.recipes` before `FD.entities`. It is schema-valid and round-trips, so it costs nothing here. Chasing it did turn up a wider pre-existing bug in that function, which I am filing separately.

Thanks for turning this around so quickly.

PR #265, opened 2026-08-25
Preserve an icon's signal type instead of re-deriving it

### LINKED ISSUE #264: Blueprint.serialize() rewrites icon signal types, changing 37.6% of corpus icons on a round trip
`Blueprint.serialize()` rewrites an icon's signal **type**, so a blueprint that
arrives with `{"type":"item","name":"wooden-chest"}` is written back out as
`{"type":"recipe","name":"wooden-chest"}`.

Measured across the committed corpus, decoding the blueprint strings directly
rather than going through the model: **359 of 956 blueprint icons (37.6%) change
type on a round trip.**

| Change | Count | Example |
| --- | --- | --- |
| `item` -> `recipe` | 334 | `rail-signal` |
| `space-location` -> `item` | 17 | `fulgora` |
| `item` -> `fluid` | 8 | `parameter-0` |

## Root cause

Two places, and the first is what makes the second necessary.

`Blueprint.ts:138` keeps only the **name** when a blueprint is parsed:

```ts
this.icons.set(icon.index, icon.signal.name)
```

`icons` is a `Map<1 | 2 | 3 | 4, string>`, so the incoming type is dropped and
cannot be preserved later.

`Blueprint.ts:886` then re-derives a type from that name alone:

```ts
const getItemTypeForBp = (name: string): SignalType => {
    if (FD.signals[name]) return 'virtual'
    if (FD.fluids[name]) return 'fluid'
    if (FD.recipes[name]) return 'recipe'
    if (FD.entities[name]) return 'entity'
    return 'item'
}
```

Three problems with that lookup:

1. `FD.recipes` is checked before `FD.entities`, and most placeable things are
   both. `wooden-chest` is an item, an entity and a recipe, so it comes back
   `recipe`.
2. `FD.items` is never consulted at all. `item` is only the fallthrough, so a
   name that really is an item gets it only by failing every other test.
3. There is no `space-location` branch, so `fulgora` falls through to `item` and
   names an item that does not exist.

## Why nothing catches it

`tests/blueprint-round-trip.spec.ts` hashes the serialized output, so the wrong
types are already baked into the committed fixture and the suite is green on
them. Nothing anywhere asserts that an icon's type survives a round trip.

## Reproduction

With `FD` loaded from the real `data.json`:

```ts
const bp = new Blueprint({
    icons: [{ index: 1, signal: { type: 'item', name: 'wooden-chest' } }],
    entities: [{ entity_number: 1, name: 'wooden-chest', position: { x: 0, y: 0 } }],
})
bp.serialize().icons
// [{ signal: { type: 'recipe', name: 'wooden-chest' }, index: 1 }]
```

## Scope

`Book` is **not** affected. It holds `icons` as a whole `IIcon[]` and passes it
through unchanged (`Book.ts:17`, `:25`, `:75`), so a book's icons survive. This
is only `Blueprint`.

This is also not the cause of the "blueprint book icons using planet names show
no icon" limitation in `CLAUDE.md`, which I checked before assuming otherwise.

---

## Update: measured what the fix needs, and one claim above was wrong

**Correction.** This issue first said "a name cannot decide this on its own",
citing `rail-signal` as both an item and a recipe. The corpus refutes that as a
practical matter. Taking all 1034 icons across blueprints and books, over 152
distinct names, **no name is ever written with two different types**. Factorio's
own output is a consistent name-to-type mapping, so re-deriving from the name is
workable rather than fundamentally broken.

The corpus cannot prove a name is *never* ambiguous, only that none of these 152
is. That is the same limit as PR #222's snapping measurement, where 325 rows
agreed and still could not establish a requirement.

### Fixing the lookup order gets almost all of it

Transcribing the candidate rule into the analysis and scoring it against what
Factorio actually wrote, before writing any code:

| Rule | Names correct |
| --- | --- |
| current (`signals, fluids, recipes, entities`, then `item`) | 55 / 152 |
| check `FD.items` **first**, then the rest unchanged | **148 / 152** |

That single reorder fixes **342 of the 359** flipped icons: all 334
`item -> recipe` and all 8 `item -> fluid`.

### The last 4 names need a list `data.json` does not have

The 4 remaining are all `space-location`: `fulgora`, `gleba`, `nauvis`,
`vulcanus` (17 icons). `data.json` exports 11 collections and **none of them
contains any planet or space location** - those four names appear nowhere in the
file, so no lookup can classify them.

The complete set at the 2.0.77 tag of `wube/factorio-data`, from the `planet` and
`space-location` prototype definitions, is 8 names:

```
aquilo  fulgora  gleba  nauvis  vulcanus
shattered-planet  solar-system-edge  space-location-unknown
```

Two ways to supply it:

- **Hardcode the list**, with a comment naming the tag it came from. There is
  direct precedent: `PositionGrid.ts` hardcodes the elevated-rail collider set
  for exactly this reason, because `data.json` does not export `collision_mask`.
- **Export it**, which means a `packages/exporter` change and regenerating
  `data.json` from a Factorio install. That overlaps #155.

**No oracle run is needed for any of this.** The corpus is ground truth for what
Factorio writes, and the planet list is a grep of `factorio-data` at the target
tag - the first two sources in CLAUDE.md's stated order, with the game never
started.

### What a fix still has to decide

- **Preserve the incoming type, or re-derive it?** Re-deriving is now measured as
  viable and is much the smaller change. Preserving is strictly safer and makes
  the ambiguity question moot, but changes the `icons` map's type and every
  reader of it.
- **`generateIcons` needs the derivation rule either way**, since it picks a name
  and has no incoming type to preserve. So the reorder and the space-location
  list are needed whichever route is taken.
- **The pinned hash in `blueprint-round-trip.spec.ts` will move.** `CLAUDE.md`
  calls that fixture a fixed point rather than a refreshable snapshot. The
  before-and-after counts above are the justification, and a new test pinning
  type preservation should land with it, since nothing asserts it today.

Found while reviewing #258, which is unrelated and correct. Its new fallback path
reaches the same function, so `captive-biter-spawner` serializes as `recipe`
rather than `entity`; that is this bug, not that PR's.

wormeyman:
Closes #264.

## The bug

An icon carries a signal type, and the blueprint it came from chose it. `Blueprint` kept only the **name** (`Blueprint.ts:138`) and re-derived the type on serialize, so a blueprint arriving with `{"type":"item","name":"wooden-chest"}` went back out as `{"type":"recipe","name":"wooden-chest"}`.

Measured over the committed corpus, decoding the strings directly: **359 of 956 blueprint icons (37.6%) changed type on a round trip.**

| Change | Before | After |
| --- | --- | --- |
| `item` -> `recipe` | 334 | 0 |
| `space-location` -> `item` | 17 | 0 |
| `item` -> `fluid` | 8 | 0 |
| **total flipped** | **359 / 956** | **0 / 956** |

## The fix

The map holds the whole `ISignal`, so a parsed icon goes back out as it came in and nothing is derived for it.

**The omitted case matters as much as the present one.** 383 of the corpus's 1034 icons omit `type` entirely, which is what Factorio does for an item. Preserving only the types that were *present* would still have rewritten more than a third of them, so the absence survives too. There is a test for exactly that.

## `generateIcons` still has to derive

It invents a name and has no type to keep. That is now `deriveSignalType`, and its order is measured rather than chosen. Scoring each candidate against what Factorio itself wrote for the 152 distinct icon names in `test-blueprints/`, **before writing any code**:

| Rule | Names correct |
| --- | --- |
| old (`signals, fluids, recipes, entities`, then `item`) | 55 / 152 |
| check `FD.items` **first** | **148 / 152** |

## The 4 it still misses are planets, and that is deliberate

`data.json` exports no planet in any of its eleven collections, so no lookup can reach one.

A first draft hardcoded the 8 `planet` and `space-location` prototype names from `wube/factorio-data` at tag 2.0.77, following the `PositionGrid.ts` elevated-rail precedent. **Measuring showed that was dead code**: `generateIcons` only ever passes an item name (`minable.result`) or an entity prototype name, and no planet is either. A planet icon only ever arrives by being *parsed*, where it is now preserved. The arm came back out, and the comment explains why, so the next reader does not add it again.

## Types

`SignalType` gains `space-location`, `asteroid-chunk` and `quality`, which `blueprintSchema.json` already accepted. That was only ever a type error rather than a runtime one, since the values arrive from `JSON.parse` - a `space-location` signal decoded and rendered fine while the type said it could not exist.

## The moved fixed point

`blueprint-round-trip.spec.ts`'s `serializedHash` moves, `825830683` -> `-1735753444`. The **shape** of that move is the argument for accepting it:

```
blueprints 367   entities 347725   tiles 232815   wires 48869   icons 956
positionChecksum -44031602   modelPositionChecksum -63968586      all held
serializedHash   825830683 -> -1735753444                         moved
```

Every count held, `icons` included, and both position checksums held, so no blueprint gained, lost or moved anything. Only the serialized JSON differs, in the one field that was being rewritten. That is the issue #40 signature, not the #186 one, and the file's note records it that way.

## Why no oracle run

The corpus **is** ground truth here: those 367 files are real Factorio exports, so whatever type the game writes is already in them. The planet list came from a grep of `factorio-data` at the target tag. Those are the first two sources in CLAUDE.md's stated order, with the game never started.

One limit worth stating: across all 1034 icons and 152 names, no name is ever written with two different types, so re-deriving *would* have worked for everything measured. The corpus cannot prove a name is never ambiguous, only that none of these is. Preserving the incoming type makes the question moot, which is why it is the route taken rather than the cheaper reorder alone.

## Verification

- `vp check` - 231 files formatted, 0 warnings, lint errors or type errors
- `vp test` - **194/194** across 16 files (6 new)
- `npx playwright test` - **198/198**
- Corpus re-measured through the model after the fix: 956 icons checked, **0 flips**

https://claude.ai/code/session_018N3pm7fQQDv6HTVz1TEpmE


wormeyman:
@claude please review this PR.

Worth focusing on, since these are the parts where I'd most expect a mistake:

1. **`Blueprint.ts` constructor** now stores `{ ...icon.signal }` rather than the name. Is the shallow copy the right depth, and is there any path where an icon signal object is mutated after being stored?
2. **`serialize()`** returns `{ ...signal }` and no longer guarantees a `type` field. 383 of the corpus's 1034 icons legitimately omit `type`, so that is intended, but check nothing downstream assumed the field was always present.
3. **`deriveSignalType`** is only called from `generateIcons`. I removed a `space-location` arm after measuring it unreachable, on the grounds that `generateIcons` only ever passes an item name (`minable.result`) or an entity prototype name. Please check that reasoning against the actual call sites rather than the comment.
4. **`setIcon` is typed `(index: 1 | 2, ...)`** where the map key is `1 | 2 | 3 | 4`. That narrowing is deliberate since `generateIcons` only writes 1 and 2, but confirm nothing else needs 3 or 4.
5. **The moved fixed point** in `blueprint-round-trip.spec.ts`. Every count and both position checksums held and only `serializedHash` moved; is that argument sound for accepting the new value?

Also flagging that `SignalType` widened by three members. It has few usages, but confirm the widening cannot make some other check non-exhaustive.

wormeyman:
Thanks - taking the minor one, and the fix turned out to be better than the suggestion.

The `as never` was a leftover from before `SignalType` widened **in this same PR**. When the helper was written, `{ type: 'space-location', name: 'nauvis' }` was not a valid `ISignal`, so the parameter was typed loosely and the whole literal cast past the checker. Now that `space-location` is a `SignalType`, the helper takes an `ISignal` and the cast goes entirely rather than narrowing to `as Partial<IBlueprint>`. The fixtures are type-checked like any other code again, which is what the cast was suppressing. `b8c605ff`.

On the verification note: that is a real limit of the review environment rather than something to work around here. The numbers in the PR body were measured locally - `vp check` clean, `vp test` 194/194, `npx playwright test` 198/198 with both dev servers up, and a throwaway probe that ran the corpus back through the model and reported 956 icons checked, 0 flips. Since the `serializedHash` argument rests on Playwright output, it is worth saying that the `e2e` job on this PR is running the same suite that produced it, so the fixed point is being checked by CI rather than only by my say-so.

PR #266, opened 2026-08-25
Give the Claude workflows a toolchain, so a review can run the gates

wormeyman:
## Why

A review that cannot run anything can only reason about the diff. That limit showed on #265, where the run said so itself:

> I didn't have `vp` or an installed `node_modules` in this environment, so I wasn't able to run `vp check`, `vp test`, or the Playwright suite myself - the analysis above is from reading the diff and tracing call sites/consumers directly rather than from re-executing the verification numbers quoted in the PR body.

It then accepted a moved `blueprint-round-trip.spec.ts` fixed point on the strength of the argument. It was right, and it was right for a reason nobody could check from inside the review.

## What

Both workflows now do what the `checks` job in `ci.yml` already does:

```yaml
- name: Set up Vite+
  uses: ./.github/actions/setup-vp

- name: Install dependencies
  run: vp install --frozen-lockfile
```

`setup-vp` writes `~/.vite-plus/bin` to `$GITHUB_PATH`, so `vp` is on PATH for the `claude-code-action` step that follows. Both caches are shared with `ci.yml`.

**Measured on this PR's own `claude-review` run**, rather than estimated: `Set up Vite+` took **6s** and `Install dependencies` **9s**, so **15 seconds** total. An earlier draft of this description said about 1m20s, which was wrong - that is the whole `checks` job including `vp check` and `vp test`, and a review only pays for those if it chooses to run them.

## Not the Playwright half

That needs browsers and both dev servers, and would put the `e2e` job's 3-4 minutes on **every** review, including the automatic one on every push. `vp check` and `vp test` cover what a reviewer most often wants to confirm, and answer in seconds once installed.

## Two things this does not settle

Stated rather than asserted, because I have not measured them:

- **`claude.yml`** places no `--allowedTools` restriction, and is the path the #265 review actually ran on. That is the case this addresses.
- **`claude-code-review.yml`** restricts `--allowedTools` to the inline-comment MCP tool. Whether that is additive to a default tool set or exclusive is untested. If it is exclusive, the install is inert there and that line needs the gates added to it explicitly. I would rather find that out from a run than guess at the argument syntax.
- **`claude.yml` also fires on issue comments**, where a toolchain is often not wanted. That cost is accepted rather than special-cased, since an `@claude` question about code is exactly when being able to run a gate helps.

## Verification

Both files re-parsed as YAML after the edit, and the step order is `Checkout -> Set up Vite+ -> Install dependencies -> Run Claude`.

The mechanical half is confirmed on this PR: in its own `claude-review` job, `Set up Vite+` and `Install dependencies` both completed successfully before `Run Claude Code Review` started. So the toolchain is present. Whether the *agent* is permitted to invoke it is the separate `--allowedTools` question above, and an `@claude` run has been asked to answer it by trying.

Note fork PRs skip `claude-review` entirely as of #263, so this only ever runs on in-repo branches - the same trust level as CI running the code.

https://claude.ai/code/session_018N3pm7fQQDv6HTVz1TEpmE

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -39,6 +39,24 @@ jobs:
         with:
           fetch-depth: 1
 
+      # A review that cannot run anything can only reason about the diff. That
+      # is a real limit and it showed: reviewing PR #265 the run said outright
+      # that it had no `vp` and no `node_modules`, so it accepted a moved
+      # `blueprint-round-trip.spec.ts` fixed point on the strength of the
+      # argument rather than by re-running the corpus. These two steps are what
+      # the `checks` job in ci.yml does, and they cost it about 1m20s with both
+      # caches warm.
+      #
+      # Deliberately not the Playwright half. That needs browsers and both dev
+      # servers, and would put the e2e job's 3-4 minutes on every review
+      # including the automatic one on every push. `vp check` and `vp test`
+      # cover what a reviewer most often wants to confirm.
+      - name: Set up Vite+
+        uses: ./.github/actions/setup-vp
+
+      - name: Install dependencies
+        run: vp install --frozen-lockfile
+
       - name: Run Claude Code Review
         id: claude-review
         uses: anthropics/claude-code-action@v1
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -30,6 +30,24 @@ jobs:
         with:
           fetch-depth: 1
 
+      # A review that cannot run anything can only reason about the diff. That
+      # is a real limit and it showed: reviewing PR #265 the run said outright
+      # that it had no `vp` and no `node_modules`, so it accepted a moved
+      # `blueprint-round-trip.spec.ts` fixed point on the strength of the
+      # argument rather than by re-running the corpus. These two steps are what
+      # the `checks` job in ci.yml does, and they cost it about 1m20s with both
+      # caches warm.
+      #
+      # Deliberately not the Playwright half. That needs browsers and both dev
+      # servers, and would put the e2e job's 3-4 minutes on every review
+      # including the automatic one on every push. `vp check` and `vp test`
+      # cover what a reviewer most often wants to confirm.
+      - name: Set up Vite+
+        uses: ./.github/actions/setup-vp
+
+      - name: Install dependencies
+        run: vp install --frozen-lockfile
+
       - name: Run Claude Code
         id: claude
         uses: anthropics/claude-code-action@v1


wormeyman:
@claude this PR is meant to give you a working toolchain. Please verify it actually did, rather than reviewing the diff.

Specifically, run these and paste the real output:

1. `vp --version`
2. `vp check`
3. `vp test`

If any of them fail or are not permitted, say exactly what happened - "not installed", "command not allowed", or the error text. A negative result here is the useful one, because it tells us whether the `--allowedTools` line in `claude-code-review.yml` is additive or exclusive, which the PR body flags as untested.

Then, if the gates do run, a short review of the diff itself.

wormeyman:
Closing. The install steps work and are not enough, and the measurement is worth keeping even though the change is not.

## What was measured

**The install half works.** In this PR's own `claude-code-review.yml` run ([32891825951](https://github.com/FactoryGameFan/factorio-blueprint-editor/actions/runs/32891825951)): cache hit on `vite-plus-Linux-0.2.9`, `VITE+ successfully installed!`, then `vp install --frozen-lockfile` added 279 packages. **6s + 9s, 15 seconds total** - not the ~1m20s this description first estimated, which was the whole `checks` job including the gates a review would only pay for if it ran them.

**Installing is not sufficient.** `vp` is refused with `This command requires approval` before it executes. That is a baseline permission gate rather than the `--allowedTools` line: the run that hit it was under `claude.yml`, which passes no `--allowedTools` at all. The decisive control is `npm`, which is preinstalled on `ubuntu-latest` and was blocked with the identical message, while `echo`, `git`, `ls`, `find`, `cat` and `node --version` all ran immediately. So it is a permission wall, not a `PATH` problem.

**No fix here can be tested before merging.** `claude-code-action` validates that the workflow file matches the default branch and skips itself otherwise:

```
##[warning]Skipping action due to workflow validation: Workflow validation failed.
  The workflow file must exist and have identical content to the version on the
  repository's default branch.
Exiting due to workflow validation skip
```

So this PR's automatic `claude-review` reported **pass** while having run no review at all - the install steps ran and the action then exited. Any change to `claude.yml` or `claude-code-review.yml` is unverifiable until it has already landed, which makes each attempt a merge-then-find-out.

## Why closing rather than iterating

Fixing it means guessing at an `--allowedTools` value, merging it unverified, and discovering on the next PR whether it worked - with the extra risk that the flag may be exclusive rather than additive, in which case adding `Bash(vp:*)` to `claude.yml` would strip the default tools it currently relies on.

Against that, the gap is self-reporting. The #265 review said plainly that it could not run the gates and was reasoning from the diff instead; this run said plainly what blocked it. And CI runs `vp check`, `vp test` and the full Playwright suite on the same commit regardless, so the coverage exists - it is just not inside the review. A reviewer that states its limits accurately is a workable reviewer.

Worth knowing if anyone revisits: **`claude-review` reporting green does not prove a review happened.** Check the job log for `Exiting due to workflow validation skip` before reading a pass as a clean review.

Issue #142, opened 2026-07-28
Use the game's real tile_width/tile_height for rails (runtime property, not an exporter omission)


wormeyman:
Split out of #133, which is closed. Item 1 there - per-rail occupancy shapes - was measured and found **not implementable as written**: occupancy is not a property of the rail, since which cells it makes unbuildable depends on the size of the box being asked about (`tools/oracle/fixtures/rail-occupancy.json`, PR #138). This is the one piece of that item that *is* well defined, and it is smaller and different in kind.

## The gap

The game publishes `tile_width`/`tile_height` on every entity prototype. The exporter emits them for **3 of 155**. So `getEntitySize` computes the footprint from `collision_box` instead, and the two disagree for every curved rail type:

| rail | game's `tile_width` x `tile_height` | editor's computed rectangle |
| --- | --- | --- |
| `curved-rail-a` | 2x4 | 2x6 |
| `curved-rail-b` | 2x2 | 2x5 |
| `legacy-curved-rail` | 4x8 | 4x4 |

Note `legacy-curved-rail` runs the other way - the editor's rectangle is **smaller** than the game's.

## What this does and does not fix

**Does**: makes the rectangle agree with what the game says the entity's tile footprint is, for every entity and not only rails. That is a straightforward correctness win over computing a number the data already contains.

**Does not**: fix occupancy. Even a correct rectangle is still a rectangle, and a curved rail's cells are not one - the measurement in #133 found 34 of 38 rail orientations blocking a cell the editor does not key and 28 of 38 keying a cell that is empty. Closing *that* needs per-pair measured rules, which is what #134, #136 and #137 each did and is the only approach that has closed anything so far.

So this is worth doing on its own terms, not as a step toward the occupancy work.

## The catch, which is the reason this is its own issue

`entity.size` does not only feed placement. `getEntityAtPosition` reads it too, so changing a footprint **moves what hovering and clicking select**. A `legacy-curved-rail` going from 4x4 to 4x8 doubles the area that selects it; `curved-rail-a` going from 2x6 to 2x4 shrinks it.

That means this is a behaviour change to the editor's input handling, not just to its placement rules, and it wants coverage on the hover side before it lands. `tests/editor-mode-input.spec.ts` has the machinery (`entityScreenPosition`, `hoveredEntityNumber`); nothing currently asserts *which* entity a click at a given offset selects.

## Work

1. Emit `tile_width`/`tile_height` from the exporter for all entities that declare them.
[...]

PR #267, opened 2026-08-25
Measure runtime rail boxes against occupancy

### LINKED ISSUE #251: data.json disagrees with the running game about every rail's collision box, and only about rails
Measured on Factorio **2.0.77**, the version this editor targets. Fixture: `tools/oracle/fixtures/rail-box-orientation.json`, probe: `tools/oracle/probe-rail-box-orientation/`. Headless create run, no player needed.

Every entity's runtime `collision_box` was dumped and compared against `packages/exporter/data/output/data.json`:

| | count |
| --- | --- |
| entities compared | 155 |
| agree to within one 1/256 step | 139 |
| disagree beyond that | **16** |
| of those 16, how many are rails | **16** |

**The 139 are not a finding.** Factorio stores positions in 1/256 of a tile, so a runtime box is the declared value snapped to that grid. The worst deviation among them is 0.00375 against a quantum of 0.00390625. Separating the two mattered: lumping them together buries 16 real differences under 139 uninteresting ones, and "152 of 155 differ" was the first, useless, version of this result.

**The 16 are every rail prototype plus its `dummy-` and `elevated-` variants.**

```
legacy-curved-rail    deviation 1.4531
  data.json [-2, -2, 2, 2]
  runtime   [-0.75, -0.546875, 0.75, 1.5976562]

half-diagonal-rail    deviation 0.3376
  data.json [-0.75, -2.236, 0.75, 2.236]
  runtime   [-0.75, -1.8984375, 0.75, 1.8984375]

legacy-straight-rail / straight-rail (+ elevated, dummy)   deviation 0.3008
  data.json [-1, -1, 1, 1]
  runtime   [-0.6992188, -0.9882812, 0.6992188, 0.9882812]

rail-ramp / dummy-rail-ramp   deviation 0.2023
  data.json [-1.8, -7.8, 1.8, 7.8]
  runtime   [-1.5976562, -7.5976562, 1.5976562, 7.5976562]

curved-rail-a / curved-rail-b (+ elevated, dummy)   deviation 0.0508
  data.json [-0.75, -2.516, 0.75, 2.516]
  runtime   [-0.6992188, -2.515625, 0.6992188, 2.515625]
```

`legacy-curved-rail` is the striking one: `data.json` says a symmetric 4x4 box, and the runtime says an asymmetric one about 1.5 wide by 2.1 tall.

## Why it matters

`getEntitySize` in `factorioData.ts:617` derives an entity's tile footprint by ceiling this box, and that footprint is what `PositionGrid` keys placement on and what `getEntityAtPosition` reads to decide what a click selects. So for rails specifically, the editor's geometry is derived from numbers the game does not agree with.

CLAUDE.md already records rails as the place where the editor's footprints are wrong in both directions, from #133 and #142, and #138 costed and abandoned per-rail measured cell sets. This adds a cause nobody had looked at: it is not only that a rectangle cannot hold a curve, it is that the box the rectangle is computed from is itself wrong for exactly these prototypes.

## What is not yet known

- **Which end is at fault.** Whether the exporter's data-stage read differs from the runtime for rails, or the exporter ran against a different Factorio build. Everything non-rail agreeing to the quantum argues against a blanket version skew.
- **Whether adopting the runtime numbers helps.** #142 is the cautionary precedent: adopting the game's published `tile_width` made agreement with measured occupancy worse in both directions, and that was only discovered by transcribing the proposed rule into the analysis and re-running it against `rail-occupancy.json` **before** writing code. Do that here too rather than assuming the game's number is automatically the better input.

## How it was found

Incidentally, while de-risking a different probe. The grid-position probe's run 2 layout depends on a `half-diagonal-rail`'s box, so the analyzer hardcodes it, so it needed a control saying "the number hardcoded here matches the game". That control fired. Running it across all 155 entities cost nothing more once the dump existed.

The same probe also found the box does **not** rotate with direction (16 directions, one box, folding to stored 0/2/4/6), and that a rail requested at (20,20) is placed at (21,21).

byalex33:
Closes #251. Extends the existing rail-box analyzer to score the captured Factorio 2.0.77 runtime boxes against the committed 38-orientation rail occupancy fixture. Only legacy-curved-rail changes editor footprint (4x4 to 2x3), producing 180 to 220 occupied-but-not-keyed cells while reducing keyed-but-empty cells from 96 to 56, so the runtime boxes are not adopted. Adds coverage and coordinate-alignment controls and records the result in the generated fixture and probe documentation. Checks: vp check --fix; vp test (194 passed); fixture rescore controls passed.

### DIFF
--- a/tools/oracle/analyze-rail-box-orientation.mjs
+++ b/tools/oracle/analyze-rail-box-orientation.mjs
@@ -27,6 +27,7 @@ import { fileURLToPath } from 'node:url'
 const HERE = dirname(fileURLToPath(import.meta.url))
 const FIXTURE = join(HERE, 'fixtures', 'rail-box-orientation.json')
 const OCCUPANCY = join(HERE, 'fixtures', 'rail-occupancy.json')
+const BASELINE = join(HERE, 'fixtures', 'entity-tile-size.json')
 const DATA_JSON = join(HERE, '..', '..', 'packages', 'exporter', 'data', 'output', 'data.json')
 
 /**
@@ -47,10 +48,22 @@ const fd = JSON.parse(readFileSync(DATA_JSON, 'utf8')).entities
 
 const box = b => [b.left_top.x, b.left_top.y, b.right_bottom.x, b.right_bottom.y]
 
-const sizeFromBox = (e, b) => ({
-    x: e.tile_width || Math.ceil(Math.abs(b[0]) + Math.abs(b[2])),
-    y: e.tile_height || Math.ceil(Math.abs(b[1]) + Math.abs(b[3])),
+/** The enclosing rectangle alone, with no declared dimension over the top of it. */
+const rawSizeFromBox = b => ({
+    x: Math.ceil(Math.abs(b[0]) + Math.abs(b[2])),
+    y: Math.ceil(Math.abs(b[1]) + Math.abs(b[3])),
 })
+
+/**
+ * `getEntitySize`: the enclosing rectangle, unless `data.json` declares a
+ * dimension, in which case the declared one wins. Reported separately from
+ * `rawSizeFromBox` because the two answer different questions - see
+ * `maskedByDeclaredDimension` below.
+ */
+const sizeFromBox = (e, b) => {
+    const raw = rawSizeFromBox(b)
+    return { x: e.tile_width || raw.x, y: e.tile_height || raw.y }
+}
 
 const swapForDirection = (type, dir, size) => {
     if (size.x === size.y) return size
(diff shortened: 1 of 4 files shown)


wormeyman (approved):
Re-derived this independently rather than reading it. My own scorer, transcribing `getEntitySize` and `processArea` from the editor source, reproduces 180/96 to 220/56 exactly. The baseline matches the `today: {missed: 180, empty: 96}` already committed in `entity-tile-size.json`, the delta is entirely `legacy-curved-rail` across its 8 orientations at plus and minus 5 cells each, the fixture is byte-identical generator output, and nothing under `packages/` changed. The verdict also survives all four reference entities, so it does not hinge on `wooden-chest`.

Two things worth adding before merge.

**Name the three rails that are actually masked.** "The other 15 differences round to the same footprint or are masked by a declared dimension" folds together two different situations. Twelve genuinely round to the same footprint. Three of them, `half-diagonal-rail` and its elevated and dummy variants, really do change, and only `data.json`'s declared `tile_height: 2` pulls them back. That masking step leans on the `tile_width`/`tile_height` read that #142 established is a centring parity rather than a size, so it is worth saying out loud rather than leaving as "masked". (Measured while reviewing: 7 of 155 entities declare one of those fields, not the 3 `CLAUDE.md` claims. That correction is in #278.)

**Add the cheap control.** `entity-tile-size.json` already holds an independently produced `today: {missed: 180, empty: 96}` against the same fixture and the same reference. Reading it and comparing would fail loudly if `sizeFromBox`, `swapForDirection` or `keyedCells` were transcribed wrong, while the hypothesis under test could still hold. That is the bar the probe README sets for a control, it is about six lines, and it is the only thing that would catch a transcription error in the new `sizeFromBox`.

Approving, since neither changes a number and the analysis is sound either way.

Issue #77, opened 2026-07-27
Enable strict: true (strictPropertyInitialization, 24 errors)


wormeyman:
Follow-on from #22, which is closed: `strictNullChecks` is on and its gate baseline has reached 0.

`strict: true` can now replace the sub-flag block in the root `tsconfig.json`. The only flag it adds that is not already on is **`strictPropertyInitialization`**.

## The count

`tsconfig.json` used to say this was "~1 error". It is not. Measured by flipping `"strict": true` after the baseline hit 0:

**24 errors across 13 files.** 23 are `TS2564` ("has no initializer and is not definitely assigned in the constructor"); one is `TS2565`.

| file | count | properties |
| --- | --- | --- |
| `UI/controls/TextInput.ts` | 5 | `state`, `_restrict_regex`, `_max_length`, `_dom_input`, `_disabled` |
| `UI/controls/Checkbox.ts` | 3 | `m_Hover`, `m_Checked`, `m_Checkbox` |
| `containers/BlueprintContainer.ts` | 3 | `_entityForCopyData`, `copyModeUpdateFn`, `deleteModeUpdateFn` |
| `UI/editors/components/Filters.ts` | 2 | `m_Filters` (TS2564 **and** TS2565) |
| `UI/controls/Slider.ts` | 2 | `p_Value`, `m_Dragpoint` |
| `containers/PaintContainer.ts` | 2 | `icon`, `_name` |
| `containers/EntitySprite.ts` | 2 | `zOrder`, `__zIndex` |
| `UI/controls/Button.ts` | 1 | `m_Data` |
| `UI/controls/Slot.ts` | 1 | `name` |
| `core/Book.ts` | 1 | `_active` |
| `containers/PaintWireContainer.ts` | 1 | `cursorBox` |
| `containers/OverlayContainer.ts` | 1 | `selectionAreaUpdateFn` |

The concentration is real and useful: **11 of the 24 are in `UI/controls/`**, which is one family of classes written to the same pattern, and 8 more are in the paint/sprite containers.

## The one that is a real question

`Filters.m_Filters` is the only property that is also **read** before assignment (TS2565), so it is the only one where the answer is about control flow rather than about an annotation. Worth starting there — the other 23 are variations on "does this genuinely have no value until later, or was the constructor just not written to say so."

## Approach

The three shapes, in the order to prefer them:

[...]

PR #268, opened 2026-08-25
Remove generated bloat and simplify the project

byalex33:
Been working on this for a few hours — hope it is not too much of a PR 😅

### Removed generated and obsolete files

- Removed 5,718 unreferenced numbered exporter artifacts (`* 2.basis`, `* 3.basis`, and numbered metadata), cutting 291.55 MiB from the checkout.
- Added ignore rules so those duplicate exporter artifacts are not recommitted.
- Removed nine completed implementation plans (7,140 lines of historical task notes).
- Removed the obsolete Pages-era `functions/corsproxy.js`; the live proxy remains in the Cloudflare Worker.
- Removed the unused worker placeholder `public/index.html`.

### Simplified application code and dependencies

- Routed assemblers, beacons, and mining drills through the existing `TempEditor` and removed three redundant editor classes.
- Replaced `file-saver` and its typings with the browser's native object-URL/download-anchor APIs.
- Replaced the single `utility-types` use with a local `FunctionKeys` type.
- Updated the lockfile; this removes three installed packages.

### Simplified tooling

- Removed the redundant type-check baseline gate, its tests, baseline file, script, CI step, and Vite test project. `vp check` already enforces strict package-aware type checking at zero errors.
- Consolidated the duplicated Rust sprite-compression implementations used by downloaded and local Factorio data.
- Added one shared Factorio oracle probe runner for isolated mod setup, execution, and dump handling.
- Migrated 17 probes to that runner, removing roughly 700 repeated lines.
- Added a small runnable self-check for the shared runner.

### Made project guidance current

- Reduced CI from 428 lines to 217 while preserving the checks, Rust builds, four Playwright shards, caching, and deployment gates.
- Updated contributing/toolchain instructions for Vite+ 0.2.9 and npm 12.
- Removed stale strict-migration and deleted-file notes from live configuration.

## Result

- 5,767 files changed
- 423 insertions, 8,740 deletions
- 291.55 MiB of duplicate generated output removed
- No new runtime dependency

## Verification

- `vp check .` — formatting, lint, and type checks pass (176 files)
- `vp test` — 172/172 unit tests pass
- `npx playwright test` — 198/198 browser tests pass
- production website build passes (3,061 assets copied)
- `cargo check --locked` passes
- `cargo clippy --locked -- -D warnings` passes
- `cargo fmt --check` passes
- oracle runner self-check and syntax checks for every oracle script pass

## Note

The `CLAUDE.md` rewrite is now isolated in #277 for independent review.

`vp install` still reports the repository's existing npm audit state (two moderate advisories and one high). I did not mix an unrelated automated dependency upgrade into this cleanup PR.

---

## Rebased on `wormeyman-space-age-support` (2026-09-02)

Merged the base branch in to clear the conflict. One file needed a manual resolution:

- `packages/website/src/index.ts` — base had extracted an `exportImage()` helper (still calling `FileSaver.saveAs`); this branch had replaced `file-saver` with a local `saveBlob`. Resolution: keep base's `exportImage()` helper and the `takePicture` action calling it, with `saveBlob` inside it instead of `FileSaver.saveAs`. `import FileSaver` stays removed; the two `FileSaver` mentions in doc comments (`ToolsPanel.ts`, `common/globals.ts`) now read `saveBlob`.

Needs a CI / local `vp check` + `vp test` + Playwright pass to confirm the merge.


wormeyman:
@claude review

wormeyman:
@byalex33 Let's put the Claude markdown file into its own pull request and then we can ship everything else in this pull request.

byalex33:
Done — restored `CLAUDE.md` here and moved its rewrite to #277 for independent review.

wormeyman:
Two blockers, both small.

1. The deleted `gate` vitest project globbed `scripts/**/*.test.mjs`, which collected two files rather than one. `scripts/localpreview.test.mjs` survives with no collector, so its 6 tests stop running. That includes `isPortFree is false for an IPv6-only listener`, which guards the macOS case where Vite binds `::1` only and localpreview reports 8080 free. Adding `scripts/**/*.test.mjs` to the surviving `unit` project's include fixes it.

2. `tools/oracle/factorio-probe.test.mjs` is collected by nothing, and it has no `import { test } from 'vite-plus/test'`, so it would register zero tests even if it were collected.

Everything risky I checked holds up. The 5,716 sprite deletions match `data.json`'s 3,060 references exactly, with no required file lost. The CI change only removes the gate step. The three editor classes render identically through `TempEditor`, because beacon and electric-mining-drill are not crafting machines, so `acceptedRecipes` is empty and the offset stays 0. The Rust consolidation is byte-identical after one identifier rename.

Two smaller things. The shared runner drops the `factorio_version` vs binary line from its no-dump error, while the migrated probe's own comment still warns about that exact silent failure. And `saveBlob` revokes the object URL on the next macrotask where file-saver waits 40 seconds, on a path with no test coverage.

wormeyman:
I pushed a merge commit to this branch (0e6753da) to bring it up to date with `wormeyman-space-age-support` and resolve the one conflict. Flagging it here since it is your branch.

The only conflict was `.github/workflows/ci.yml`. I resolved it by taking the base file and applying this branch's two real edits to it, rather than taking either side whole.

The reason: this branch's copy of `ci.yml` predates two changes on base.

- #305 added the `changes` job and the `needs:` / `if:` gating that skips web or rust work on PRs that cannot affect them. That cut an exporter-only PR from about 23 minutes of runner time to about 6.5.
- #316 moved the CI reasoning into `.github/workflows/README.md` and kept 7 short guards inline, sited where a reader is about to undo the thing the guard protects.

This branch deletes every comment in the file and has no `changes` job. Both look like artifacts of the older base rather than intent, so taking this branch's file whole would have silently reverted both of those. If either was actually deliberate, say so and I will redo the resolution.

What this branch wanted in `ci.yml` is two edits, and both are applied: the `Type-check gate` step is gone, and the vitest step is renamed to `Unit tests`. Against base that is 7 lines removed and 1 renamed.

Removing the gate is a deliberate maintainer decision, not an accident of the merge. The baseline has been back at 0 since `strict: true` landed under #77, so the gate is redundant until the next flag flip. The strict-ratchet recipe it existed for lives in `.github/workflows/README.md` and survives untouched; a follow-up PR rewords that section to read as history rather than as a live job.

Two things from the earlier review are still open, and both will land in that follow-up rather than as more commits on your branch:

1. `scripts/localpreview.test.mjs` is orphaned. Deleting the `gate` vitest project removed the only `include` matching `scripts/**/*.test.mjs`, and the surviving `unit` project globs only `tests/**/*.test.ts`. Its 6 tests stop being collected, silently. The fix is one line in `vite.config.ts`.
2. `tools/oracle/factorio-probe.test.mjs` is collected by nothing and declares no tests - it is top-level `assert` calls, so it would register zero tests even if a glob picked it up. Since it covers the new shared `factorio-probe.mjs` runner that 17 probes now depend on, the follow-up converts it to real tests rather than deleting it.

Separately, the WebKit question on `saveBlob` is settled and is not a blocker. Measured on macOS 26.6.2 with Safari 26.6.2 and Safari Technology Preview 27: the immediate revoke and file-saver's 40 second revoke produce byte-identical files, and the immediate revoke passed 5 of 5 runs. Chrome 152 passes both ways too. file-saver's "Safari doesn't allow downloading of blob URLs" comment does not describe current Safari, so dropping the FileReader fallback is safe on desktop. iOS Safari and macOS WKWebView are still unmeasured, and the export path is not mobile-gated, so that stands as a known limit rather than a tested one.

Thank you for the cleanup - 291.55 MiB of duplicated `.basis` artifacts, and the check that nothing references a numbered name held up under review.

PR #276, opened 2026-08-27
Pass a factorioprints API url through untouched, as factorio.school already is

motlin:
Factorio Prints is about to link every blueprint in a book straight into this editor, the way factorio.school already does. Those links address one blueprint inside a book by its position:

```
https://fbe.factorygamefan.com/?source=https://factorioprints.xyz/api/blueprintData/<sha>/position/1.0
```

`factorioprints.xyz` is the same API as `www.factorio.school`, under a factorioprints domain — same paths, same responses.

Today that URL loads nothing. The switch in `bpString.ts` keys on the first label of the hostname, so `factorioprints.xyz` takes the firebase arm and is rewritten to `blueprints/blueprintData.json` — `pathParts[1]` of an API path is `blueprintData`, not a blueprint key. And even read correctly, the firebase record only ever holds the whole book, so the position would be dropped.

`factorio.school` already has the fork this needs: an `/api/` path is passed through untouched and read as text. This does the same inside the `factorioprints` arm, which is why the two arms now read alike.

Also adds `factorioprints.xyz` to the corsproxy allowlist. The catch-all would relay it anyway, so nothing depends on this — it is the same reasoning that already puts bare `factorio.school` on the list: a pass-through arm's host appears nowhere in `bpString.ts` as a literal, so the "covers every host bpString.ts builds a URL for" scan cannot see it, and the entry is what says the host is expected rather than merely tolerated.

Note on the other domain: `factorioprints.com` is the site, not the API — `factorioprints.com/api/...` answers the SPA's `index.html` at status 200. It is deliberately not on the allowlist, and the `/view/` shape it does serve keeps going to firebase exactly as before.

## Tests

- `tests/blueprint-sources.spec.ts` — a factorioprints API url is passed through unchanged and read as text. Checked red: without the change it fails on the rewritten target.
- `tests/corsproxy.test.ts` — the hosts the pass-through arms fetch are allowlisted, in the other direction from the existing scan.

`vp check`, `vp test` (221) and the full `blueprint-sources` spec (12) pass locally.

## Checked against production

```
$ curl -s -o /dev/null -w '%{http_code}\n' \
  'https://fbe.factorygamefan.com/corsproxy?url=https%3A%2F%2Ffactorioprints.xyz%2Fapi%2FblueprintData%2F9030bab1975b1b0b0d07de628f0b032286394e37%2Fposition%2F1.0'
200
```

That sha is a real book whose entry 1 is a nested book, so `1.0` exercises the nested path. The proxy already relays it; only the editor-side rewrite is in the way.

### DIFF
--- a/packages/editor/src/core/bpString.ts
+++ b/packages/editor/src/core/bpString.ts
@@ -298,6 +298,22 @@ function getBlueprintOrBookFromSource(source: string): Promise<Blueprint | Book>
                         r.text()
                     )
                 case 'factorioprints':
+                    /*
+                        The same fork as the factorio.school arm below, and for
+                        the same reason. `factorioprints.xyz` is the same API
+                        under a factorioprints domain, and a link to one
+                        blueprint *inside a book* is an
+                        `/api/blueprintData/<sha>/position/<i>` URL. The firebase
+                        record holds the whole book and nothing else, so
+                        rewriting to it would silently drop the position and open
+                        the book instead of the blueprint that was linked -
+                        `pathParts[1]` of an API path is `blueprintData`, not a
+                        blueprint key, so in practice the fetch fails outright.
+                    */
+                    if (pathParts[0] === 'api') {
+                        return fetchData(url.href).then(r => r.text())
+                    }
+
                     return fetchData(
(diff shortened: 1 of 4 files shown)


wormeyman:
@claude review

Issue #231, opened 2026-08-17
F.CreateIcon throws on planet icon names, taking down whatever called it


wormeyman:
`F.CreateIcon` ends in a bare `throw` for any name that is not an item, fluid, recipe, signal or inventory group. Space Age planet names are none of those, and blueprints in the wild carry them as icons.

Measured against `packages/exporter/data/output/data.json`: `vulcanus`, `fulgora`, `gleba` and `nauvis` appear in `FD.items`, `FD.fluids`, `FD.signals`, `FD.recipes` and `FD.inventoryLayout` a combined zero times.

Measured against the committed corpus: 19 icon references use those four names, 17 of them at blueprint level, across 5 of the 12 files.

```
JEPAKAZOL/vulcanus-starter-mk2.txt      5 blueprints, each with a vulcanus icon
JEPAKAZOL/fulgora-mall-4-planets.txt    fulgora
JEPAKAZOL/gleba-mall-5-planets.txt      gleba
JEPAKAZOL/nauvis-starter-bot-rush.txt   nauvis
JEPAKAZOL/space-platform-factory.txt    fulgora
```

This is recorded in CLAUDE.md as a limitation ("Blueprint book icons using planet names show no icon"), which is accurate for the read paths that already catch. The problem is the read paths that do not.

## Why it matters now

Two open PRs add unguarded calls on a path with no `try`/`catch` above it.

**#222**, `BlueprintInfoEditor.ts:40`. `BlueprintIconSlot.updateContent` calls `F.CreateIcon(name)` directly. Loading `vulcanus-starter-mk2.txt` and clicking the corner button:

```
Error: No item, fluid, recipe, signal or inventory group named vulcanus
    at CreateIcon (functions.ts:147)
    at BlueprintIconSlot.updateContent (BlueprintInfoEditor.ts:35)
    at new BlueprintIconSlot (BlueprintInfoEditor.ts:22)
    at new BlueprintInfoEditor (BlueprintInfoEditor.ts:79)
```

The dialog never opens, and it throws again on every later click.

There is a second effect that outlives the click. `Dialog`'s constructor runs `Dialog.s_openDialogs.push(this)` at line 37, before the subclass body throws, so each failed open leaves an entry nothing removes. `openDialogCount` reads `dialogsContainer.children.length` and cannot see it, but `Dialog.anyOpen()` can. Measured with a control:

| | press E | openDialogCount |
| --- | --- | --- |
[...]

PR #277, opened 2026-08-28
Refresh CLAUDE.md and address the #278 review

byalex33:
## What changed

Replaced the 996-line historical `CLAUDE.md` with a current development guide, and addressed the review in #278.

Split from #268 so the documentation rewrite can be reviewed independently from the mechanical cleanup.

### The guide

- New `CLAUDE.md` — repository layout, setup, architecture, invariants, sprite-data shapes, testing, the oracle workflow, dependencies, deployment, known limits.

### #278 group 1 — wrong statements in the new file

- Recipe ingredients/results: the real trap is that an empty Lua list encodes as `{}` (survives `!== undefined` and `?? []`, then throws in the first `for-of`), not element shape.
- Planet icons: `F.CreateIcon` ends in a bare `throw`, so an uncaught path loses the whole blueprint, not just the icon (#231).
- Rail placement is wrong in **both** directions — 24 measured refusals the game accepts (#133), not only "more permissive".
- `getDirName`: named `railgun-turret`, pinned failing in `tests/__fixtures__/sprite-data.json` because the corpus places it at directions 2 and 14.
- Sprite filename prefixes: five, not four — added `__elevated-rails__`.
- Repository layout: added `docs/superpowers/specs`.

### #278 group 2 — findings that only lived in the old file

Moved to a comment at the code site:

- `History.updateMap` — `undefined` means "delete this key", and `Action` swaps old/new on undo, so the callback can see `undefined` at either end.
- `WireConnectionMap.set` — a self-connection is indexed once, not twice.
- `EntitySprite.getParts` — `zOrder` is assigned from the pre-skip index.
- `controls/functions.ts` `CreateIcon` — the bare `throw` and what it costs an uncaught caller (#231).
- New `spriteDataBuilder.ts` file header — the `draw_*` pattern table, plus the two Space Age hazards (`e.name` switches need a `default`; layer keys are per entity, not per type).

Kept in `CLAUDE.md`: a Dependencies section (Renovate vs Dependabot, the `undici` hold, do-not-`npm audit fix`, `ajv`), the `vp check` rationale, and two method rules.

### Also

- `CONTRIBUTING.md` — the piped `curl … | bash` install form does not pass `VP_VERSION` to the installer; replaced with the file-download form. `npm ^11` → `^12` to match the root.
- `tools/logo/README.md` — cited `docs/superpowers/specs/2026-08-05-hard-fork-design.md`, which only exists on closed branch/PR #192; now points at PR #237 and #192.

## Ordering

**Merge after #268.** The oracle section documents `tools/oracle/factorio-probe.mjs`, which #268 adds. Per #278 item 1, "merge #268 first" is the resolution.

## Verification

Docs plus code comments; the only executable change is the merge with the base branch. `vp check` / `vp test` / Playwright should confirm on CI.


wormeyman:
I filed #278 with what this needs before merge: 7 statements in the new file that are wrong, and 14 findings from the old file that have no other home in the repo. Most of those 14 are one code comment each rather than prose to put back into `CLAUDE.md`.

The trim itself checks out. The oracle method, the dependency holds and the per-spec rationales really are duplicated at the code site, and four things in the old file were already stale.

One ordering note: this needs to land after #268. The new file documents `factorio-probe.mjs`, which only exists on that branch.

wormeyman:
Alex, I pushed a commit to this branch to unblock it. Shout if you'd rather I hadn't.

The only thing tying this PR to #268 was line 189 naming `factorio-probe.mjs`, which arrives with that PR. I reworded the paragraph to describe what a probe does rather than which module does it, so it stays accurate both before and after #268 lands.

I checked every backtick-quoted path in the new guide against `wormeyman-space-age-support`. That was the only dangling one. The other four that do not resolve are the repo name, `packages/exporter/.env`, `packages/website/dist` and `~/.vite-plus/bin`, all of them fine.

The new wording is checked against the probes already on base: `probe-rail-placement`, `probe-zoom-limits` and `probe-entity-tile-size` each call `mkdtempSync` into `tmpdir()`, write their own `info.json`, read `FACTORIO_BIN` and end with `error("DUMPED-OK")`.

`vp check` passes locally. With that line gone the merge order no longer matters, so this can go in without waiting on #268 and its 5,767 files. The "Ordering" section of the PR description is now stale.

Issue #109, opened 2026-07-27
Editing factorioData.ts with the dev server live empties FD, and the suite then fails somewhere unrelated


wormeyman:
Editing `factorioData.ts` with the dev server running empties `FD` under the live app, and the Playwright suite then fails somewhere unrelated with an error that reads exactly like a regression.

## What happens

`FD` is a module-level object that starts empty and is filled in by `loadData` at startup:

```ts
const FD = {} as FactorioData

export function loadData(str: string): void { /* fills in all twelve properties */ }
```

`packages/editor/src/core/factorioData.ts:867`

Vite's HMR re-evaluates the module on edit, so `FD` is rebound to `{}` - but `loadData` is called once during startup, from code that HMR does not re-run. The app keeps going with an empty `FD`.

## Why it earns an issue rather than only a note

The failure surfaces nowhere near the edit. Observed while working on #55: one spec failed that I had not touched, with

```
TypeError: Cannot read properties of undefined (reading 'requester-chest')
    at Ajv.validate (bpString.ts:34)
```

which is an AJV custom keyword reading `FD.entities`. Nothing in the message mentions `FD`, `loadData` or HMR. Restarting the servers took the same suite from `1 failed` to `97 passed` with no code change.

The cost is not the restart, it is the time spent debugging a regression that isn't one - and the risk of "fixing" real code to chase it. The failure is silent in the browser too: the app renders an empty editor rather than erroring, so a manual test after an edit can look merely blank.

## What exists today

A note in `CLAUDE.md` (landed in #106) telling whoever reads it to restart the servers. That helps a reader who has already suspected the cause; it does nothing for the person watching an unrelated spec fail.

## Possible fixes, roughly in order of confidence

1. **Force a full reload for this module.** `import.meta.hot.invalidate()` (or `accept(() => location.reload())`) in `factorioData.ts` makes an edit reload the page, which re-runs startup and so re-runs `loadData`. Smallest change, and it targets the exact module with the once-only init. There are no `import.meta.hot` uses anywhere in `packages/` today, so this would be the first.
2. **Make the emptiness loud.** A getter or an explicit `loaded` flag that throws `FD has not been loaded - restart the dev server` on read before `loadData` runs. Turns a confusing `TypeError` deep in AJV into a sentence naming the cause. Complementary to (1), and it would also catch any other route to an unloaded `FD`.
[...]

PR #302, opened 2026-09-02
Keep the Playwright test API out of the production bundle

### LINKED ISSUE #292: The Playwright test API ships in the production bundle
`window.__fbe_test` is assigned unconditionally at `packages/website/src/index.ts:280`, so the whole Playwright test API is in the bundle deployed to fbe.factorygamefan.com.

Measured rather than assumed: a local production build (`packages/website/dist/assets/index-*.js`) contains `__fbe_test` along with `spriteDataTally`, `recipeShapeTally`, `overlayInfoTally` and `paintPreviewTally`. There is no `import.meta.env` or `NODE_ENV` guard anywhere in the repo, and `packages/website/vite.config.js` has no `define` or plugin that would strip it.

Found by @motlin while reviewing #291. He used the hook to check that PR's behaviour against production, which is what surfaced it.

This is a known state rather than a new hazard, and that is worth writing down before anyone reaches for a fix. The comment at `packages/website/src/index.ts:435` already says the hook is assigned unconditionally and "sits on `window` at fbe.factorygamefan.com for any script on the page to call", and `exportGuardResult` exists in its current form because of it: the earlier version called `exportString()` and `exportImage()`, so any script on the page could trigger a real clipboard write and a real PNG download. The members have been read once already with production exposure in mind.

What is left is whether the surface should ship at all. It is about 358 lines and 35 members, including mutators (`loadBp`, `setEntityFilters`, `setWagonInventory`, `selectBookIndex`) and four tally functions that exist only to feed fixture specs.

Two things to settle before guarding it.

Every spec waits for `window.__fbe_test` and reads its arrival as "the editor is ready" (#109). Whatever guard goes in has to keep that true for whatever the `e2e` job builds. That job runs `npm run localpreview`, which is the dev server, so an `import.meta.env.DEV` guard would probably be fine, but that wants checking rather than assuming.

Whether the tally members retain code that would otherwise tree-shake away. That decides whether removing them is a bundle-size win or only tidiness.

byalex33:
## What

`window.__fbe_test` - and the ~360-line test API object behind it - was assigned unconditionally at the end of startup in `packages/website/src/index.ts`. Every `vp build` shipped the whole thing to fbe.factorygamefan.com: the mutators (`loadBp`, `setEntityFilters`, `setWagonInventory`, `selectBookIndex`), the four fixture tally functions, `ThrowingDialog`, and the digest helpers.

## Fix

Gate the assignment on `import.meta.env.DEV`:

```ts
.then(() => {
    startupFinished = true
    if (import.meta.env.DEV) {
        ;(window as any).__fbe_test = testApi
    }
})
```

- **Production**: Vite statically replaces `import.meta.env.DEV` with `false`, the branch is dropped as dead code, and Rolldown then tree-shakes `testApi` and everything only it references (`ThrowingDialog`, `spriteDataDigest`, `DIGESTED_FIELDS`, `entityOf`).
- **e2e**: the `e2e` CI job serves the app with `vp dev` (via `npm run localpreview`), where `import.meta.env.DEV` is `true`, so `window.__fbe_test` still appears and still means "the editor is ready" - the invariant `tests/test-hook-readiness.spec.ts` guards (issue #109).
- `startupFinished` stays **outside** the guard - the `copy`/`paste` clipboard listeners read it in production too. Inside, a production build would pin it `false`, and Ctrl+C/Ctrl+V on the canvas would silently stop working with no spec able to catch it.

`import.meta` requires `module` at es2020+, and the root tsconfig is on `ES6`, so `packages/website/tsconfig.json` gets `"module": "ESNext"`. That is safe because the root sets `moduleResolution: bundler` explicitly - raising `module` alone would otherwise fall back to `Classic` resolution and break every import in the package. Measured with `tsc --showConfig`: the resolved config differs from the base by that one line and nothing else.

## What only tree-shakes locally

Only the bindings declared inside `index.ts` disappear - `testApi`, `entityOf`, `ThrowingDialog`, `spriteDataDigest`, `DIGESTED_FIELDS`. The eight imports from `@fbe/editor` (`OverlayContainer`, `EntityContainer`, `EntitySprite`, `EntityInfoPanel`, `getSpriteData`, `SPRITE_GENERATION_FAILED`, `Dialog`, `FD`) stay in the bundle, because the editor's own production code still references those modules. Nothing gets smaller for them.

## On the tally members

The issue asks whether removing the four tally functions saves bundle size or is just cleanliness. It's moot: the guard already strips them along with the rest of `testApi`. No separate removal needed.

## Verification

Re-measured on the current base (the first numbers here predated #268, which moved the baseline):

- `vp check .` - clean
- `vp test` - 212 passed in 18 files. This PR adds and removes no tests; the earlier "225" was from before #268/#319 reshuffled test collection.
- `npx playwright test --list` - 243 tests in 52 files
- Production build, A/B on the same base: main chunk **842.48 kB -> 838.81 kB**, a **-3,670 byte** delta (gzip **235.32 -> 233.91 kB**, -1.41 kB). Every other emitted chunk is byte-identical except the two that embed the main chunk's content hash.

### Bundle greps

`__fbe_test`, `spriteDataTally`, `recipeShapeTally`, `overlayInfoTally`, `paintPreviewTally`, `throwingDialogAttempt`, `setWagonInventory`, `exportGuardResult`, `selectBookIndex`, `setEntityFilters` all go **1 hit -> 0** across `dist/assets/*.js`.

`DIGESTED_FIELDS`, `spriteDataDigest` and `ThrowingDialog` are **not** usable as evidence and an earlier version of this body cited them in error. They are module-local bindings, so oxc renames them in every production build - they are 0 in the baseline too, which is the minifier rather than tree-shaking. String literals discriminate instead, since no minifier renames them:

| probe | this PR | baseline |
| --- | --- | --- |
| `deliberate: a dialog constructor that throws after super()` | 0 | 1 |
| `never shown` (a `DIGESTED_FIELDS` entry) | 0 | 1 |

Confirmed the same way against `--minify false` builds of both sides, where the identifiers survive: the dropped code is present in the baseline and absent here.

Closes #292

### DIFF
--- a/packages/website/src/index.ts
+++ b/packages/website/src/index.ts
@@ -452,8 +452,13 @@ const testApi = {
      * whatever they returned, which mirrors the guard only for an *empty*
      * blueprint; for a loaded one it silently performs the real action -
      * an actual clipboard write and an actual PNG download - every time it
-     * is called. Harmless now that the API is dev-only (issue #292), but a
-     * guard predicate should still read state rather than change it. See
+     * is called. The gate below removes one half of that reach: this no
+     * longer sits on `window` at fbe.factorygamefan.com for any script on the
+     * page to call (issue #292). It does not remove the other half. The specs
+     * run against `vp dev`, where the hook is present, so a mutating predicate
+     * would still do a real clipboard write and a real PNG download on every
+     * call inside a Playwright run. What makes this safe is that it reads
+     * state instead of changing it, not the gate. See
      * tests/quick-actions.spec.ts.
      */
     exportGuardResult: () => ({ exportString: !bp.isEmpty(), exportImage: !bp.isEmpty() }),
--- a/packages/website/src/index.ts
+++ b/packages/website/src/index.ts
@@ -70,9 +70,18 @@ let book: Book | undefined
 
 /*
     Whether startup has run to the point where `bp` and the editor's `G.UI`
-    both exist - set at exactly the moment `__fbe_test` goes on `window`, and
-    for the same reason its own comment gives. Read by the clipboard listeners
-    below; see clipboardShortcutsBelongToCanvas for why they need it.
+    both exist - set in the same `.then` that would put `__fbe_test` on
+    `window`, and for the same reason its own comment gives. Read by the
+    clipboard listeners below; see clipboardShortcutsBelongToCanvas for why
+    they need it.
+
+    The two used to be one statement. Since #292 the hook is behind
+    `import.meta.env.DEV` and this flag is not, so in a production build this
+    is set and `__fbe_test` is absent. **Do not collapse them back.** Reading
+    `window.__fbe_test !== undefined` instead of this flag would leave the
+    clipboard listeners permanently disabled in production and nowhere else,
+    which is the #109 Ctrl+C crash again in the one environment no spec
+    covers - the whole suite runs against `vp dev`, where the hook is there.
 
     Measured before it existed: with the canvas focused during a cold data.json
     load, Ctrl+C threw an uncaught "Cannot read properties of undefined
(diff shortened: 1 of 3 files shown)


wormeyman:
@claude review

PR #316, opened 2026-09-02
Move the CI reasoning into a doc and leave guards at the traps

wormeyman:
`ci.yml` was 522 lines, 280 of them comments and blanks, and that prose was the
only record of several expensive measurements. #268 wants to strip 212 of those
lines. Moving the reasoning somewhere better first makes that strip a deletion
of a duplicate rather than of the original.

## What this does

- Adds **`.github/workflows/README.md`** (323 lines): the long-form reasoning for
  every job - what was measured, what was tried and rejected, and the parts that
  look simplifiable but are not. GitHub renders a README in that directory
  listing, and Actions only parses `.yml` there, so it is discoverable and inert.
- Slims **`ci.yml` from 522 to 321 lines**, done programmatically rather than by
  retyping. Comments *inside* `run:` block scalars are shell script, not prose,
  and are preserved byte for byte.
- Keeps **7 short guards inline**, sited where a reader is about to undo the
  thing the guard protects: the `changes` job's job-level `if:`, the type-check
  gate, clippy's `-D warnings` and Linux gating, `rust-windows` being a native
  runner, the Playwright sharding lever, the chromium library assertion, and the
  `deploy` double gate. A pointer alone is too weak at those seven spots.
- Adds one line to `CLAUDE.md`'s repository layout list pointing at the new doc.

## Verification

- **No behaviour change.** Strip comments and blank lines from the old and new
  `ci.yml` and diff: identical, **242 code lines on both sides**.
- **actionlint 1.7.7 with shellcheck 0.10.0 is clean** on the new `ci.yml` and on
  all four workflows. It is also clean on the `da660804` baseline, so this is a
  like-for-like result rather than a newly passing gate.
- `ci.yml` still parses to the same six jobs: `changes`, `checks`, `deploy`,
  `e2e`, `rust`, `rust-windows`.
- `.github/workflows/README.md` is committed 100644; it was 100755 in the first
  draft, which was an artifact rather than intent.

**`vp check` was not run locally.** The Vite+ toolchain is not installed on the
machine this was prepared on, and this commit touches only `ci.yml`, a new
markdown file and one markdown line, so there is no TypeScript, Rust or config
for it to check. CI runs the real gate on this PR.

## Three figures were stale and are re-measured

Re-measured from the GitHub API on 2026-09-02:

| figure | old comment | measured |
| --- | --- | --- |
| `checks` duration | 46s | **median 37s**, n=11, range 31-42s |
| `changes` job | ~5s | **median 3s**, n=5, max 7s |
| Playwright shard leg | 213/225/250s (n=3) | **median 226s**, n=44, range 170-337s |

The doc keeps both the median and the range, because the spread is the point.

## Two claims were wrong and are corrected rather than copied across

- **The skipped-job argument does not hold for a matrix job.** "A skipped
  workflow never reports where a skipped job does" is true for `checks`, `rust`
  and `rust-windows`, but a skipped matrix job never expands its matrix.
  Measured on #315: `e2e` reported as a single check under the literal name
  `Playwright ${{ matrix.shard }}/${{ strategy.job-total }}`, and `Playwright
  1/4` through `4/4` did not appear at all. Harmless today, but a required shard
  name would block merges forever.
- A comment claimed `deploy` is gated on `checks` alone when it is
  `needs: [checks, e2e]`.

## Relationship to #268

No textual conflict - #268 is `MERGEABLE`/`CLEAN` against `da660804`. It will
need a rebase and a re-read once this lands, since it strips comments it has
never read.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,323 @@
+# Why CI looks like this
+
+`ci.yml` is the only workflow that gates merges and deploys. This file holds the
+reasoning behind it: what was measured, what was tried and rejected, and which
+parts look like they could be simplified but cannot.
+
+The workflow itself keeps only short guards, placed where someone is about to
+undo the thing the guard protects. Everything longer lives here.
+
+**Every number below names how and when it was measured.** If you repeat one,
+re-measure it first. Two figures in this repo went stale without anyone
+noticing, and both were being quoted as settled fact.
+
+## The job graph
+
+```
+changes ──┬── checks ─────────┐
+          │                   ├── deploy
+          ├── e2e (4 shards) ─┘
+          ├── rust
+          └── rust-windows
+```
+
+`deploy` runs only on a push to `wormeyman-space-age-support` or a manual
+dispatch with the deploy input set. It never runs on a pull request.
+
+Merging to that branch ships to https://fbe.factorygamefan.com, so a merge is a
+release.
+
+## `changes`: skip the jobs that cannot see the change
+
+Every pull request used to run all seven jobs. A Cargo.lock-only crate bump
+spent about 23 minutes of runner time, 17 of it on four Playwright shards that
+cannot observe a Rust change. Renovate makes that the common case rather than a
+corner case: 14 of the 20 bot pull requests before 2026-09-01 were single Rust
+crate bumps.
+
+The job lists the files a pull request touches and sets two outputs. `checks`
+and `e2e` wait on `web`. `rust` and `rust-windows` wait on `rust`.
+
+Measured on PR #315, an exporter-only diff, 2026-09-02: 6.5 minutes of runner
+time against about 23 before.
+
+### Why a job-level `if:` and not `on.pull_request.paths`
+
+A workflow-level path filter skips the whole run, so the pull request shows no
+CI at all. That is indistinguishable from CI being broken, and it would block
+any check that is ever marked required, because a skipped workflow never
+reports.
+
+A skipped job does report. `checks`, `rust` and `rust-windows` all appear greyed
+out rather than missing.
+
+**One exception, and it is why this section is longer than one line.** A skipped
+matrix job never expands its matrix. Measured on PR #315: `e2e` reported as a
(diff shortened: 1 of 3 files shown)


wormeyman:
Two updates since this opened.

**Added `907962a1`: the documented Vite+ pin moves to 0.3.0.** #310 merged as
`99895b08`, but Renovate's customManager only rewrites
`.github/actions/setup-vp/action.yml`, so `CLAUDE.md` was left telling a reader
to install 0.2.9 against a 0.3.0 lockfile.

The same commit records that **`VP_HOME` stops being a no-op at 0.3.0**. A
default install now follows the XDG layout into `~/.local/share/vite-plus/bin`,
which would leave the `~/.vite-plus/bin` PATH line in that same recipe pointing
at nothing, so `vp` reads as missing rather than misplaced. That is the same
reasoning already written into `setup-vp/action.yml`; it just was not in the
recipe a human follows.

Verified locally after the merge: `vp v0.3.0`, all three `package.json` pins at
0.3.0, `vp check .` green (244 files formatted, 193 checked, 3.1s), install
stayed single-root with no XDG directory created.

**This conflicts with #314 on `CLAUDE.md`.** #314 rewrites the line

```
- `docs/superpowers/specs` — design documents for larger past changes
```

and this branch inserts a new `.github/workflows` entry immediately after it, so
the two hunks overlap. There is no semantic disagreement - the resolution is to
keep #314's reworded `docs/superpowers` line and this branch's added
`.github/workflows` line.

Suggested order: **land #314 first, then rebase this branch.** #314 is the older
and larger of the two, and it is the one whose content was measured against the
code. Rebasing this branch afterwards is a two-line resolution.

wormeyman:
Rebased onto wormeyman-space-age-support at 3c964cdb, after #314, #308 and #307 merged. The CLAUDE.md conflict is resolved as written up above: both lines kept, the reworded docs/superpowers line from #314 and the added .github/workflows line from this PR. The .github/workflows directory is byte-identical to the pre-rebase 907962a1, so the behaviour-identical verification on ci.yml still holds. Re-running the full gate.

PR #323, opened 2026-09-02
Verify the production test-API boundary and document public source maps

### LINKED ISSUE #322: Nothing verifies the test API stays out of the production bundle
#302 keeps `window.__fbe_test` out of production builds. Nothing checks that it worked, and nothing can.

The e2e suite cannot test a production build **by construction**: all 52 specs in `tests/` wait on `window.__fbe_test`, which after #302 exists only under `vp dev`. `tests/test-hook-readiness.spec.ts` guards ordering (#109) and runs against the dev server like the rest, so it passes unchanged either way.

The build path has no coverage either. `.github/workflows/ci.yml` runs `npm run build:website` in the `deploy` job, which is gated on `e2e` having already passed and ends at the wrangler action - no post-deploy smoke test. So a production-only regression has zero automated coverage in either direction:

- if the guard were reverted, the test API would ship again and nothing would fail
- if a `define` change or a DCE bug stripped more than intended, nothing would fail either

The second is the more expensive one. #302's own review turned up a related trap: three of the symbols originally cited as evidence (`DIGESTED_FIELDS`, `spriteDataDigest`, `ThrowingDialog`) are module-local bindings that oxc renames in every production build, so grepping for them gives 0 hits whether or not tree-shaking fired. Any future check needs probes that discriminate - string literals work, since minifiers do not rename them.

## Suggested shape

One spec, run against a built and previewed bundle rather than the dev server, asserting:

- the canvas renders and the editor works **without** `window.__fbe_test`
- `window.__fbe_test === undefined`

It needs a build step and a preview server the `e2e` job does not currently run, so it is not free. Worth weighing against a cheaper alternative: a unit-level assertion over `dist/assets/*.js` after a build, checking the two string-literal probes above are absent. That would need the build in CI but no browser.

Note issue #321 - a preview server currently binds the same port as the specs' default `baseURL`, which any such spec would have to account for.

### LINKED ISSUE #328: The deployed sourcemap carries the test API source that #302 removed from the bundle
#302 moved `window.__fbe_test` and the ~400-line test API behind
`import.meta.env.DEV` so a production build folds the guard to `false` and
tree-shakes it. That works: the executable bundle genuinely does not carry it,
and nothing is reachable at runtime.

The source text still ships, in the sourcemap.

`packages/website/vite.config.js` sets `build: { sourcemap: true }`, and
`packages/worker/src/index.ts` hands everything that is not `/` or `/corsproxy`
straight to `env.ASSETS.fetch(request)` with no extension filter. So the maps in
`packages/website/dist/assets` are served, and they carry `sourcesContent`.

## Measured on the live site

```
$ curl -sS https://fbe.factorygamefan.com/ | grep -o '/assets/index-[A-Za-z0-9_-]*\.js'
/assets/index-CbjUmkJf.js

$ curl -sS -o /dev/null -w '%{http_code} %{content_type} %{size_download}\n' \
    https://fbe.factorygamefan.com/assets/index-CbjUmkJf.js.map
200 application/json 3432114

$ curl -sS https://fbe.factorygamefan.com/assets/index-CbjUmkJf.js.map \
    | grep -o -e '__fbe_test' -e 'setWagonInventory' -e 'spriteDataTally' | sort | uniq -c
   5 __fbe_test
   1 setWagonInventory
   3 spriteDataTally
```

The entry chunk emits a `sourceMappingURL` line pointing at it, so a browser
devtools session finds it without guessing.

Same result against the committed `dist` locally: all seven of the strings
`tests/production-bundle.test.ts` scans for are absent from `assets/*.js` and
present in `assets/index-CbjUmkJf.js.map`.

## Why nothing catches it

`tests/production-bundle.test.ts` (#323) filters `name.endsWith('.js')`, which
correctly excludes `foo.js.map`. That is right for the question it asks - "did
the code ship" - but it means the scan is blind to this by construction, so the
gate cannot start failing on its own.

## What is actually wrong

The runtime guarantee holds and I am not claiming a vulnerability. Two narrower
things are wrong:

1. `packages/website/src/index.ts:296` says the test API "never reaches the
   production bundle at fbe.factorygamefan.com". At the level of shipped bytes
   that is false.
2. Whatever we decide, the decision is currently implicit. Nothing records that
   shipping the maps is a choice rather than an oversight.

## Options

Either close the gap or narrow the claim:

- **Close it.** Switch to `build: { sourcemap: 'hidden' }`, which still emits the
  maps (so they can be uploaded to an error reporter if we ever want one) but
  stops emitting the `sourceMappingURL` comment, and stop deploying them.
  Widen #323's scan to `*.js.map` so the property is then pinned in both
  directions.
- **Keep them and say so.** Shipping sourcemaps is a normal choice for a project
  whose source is public anyway. Then reword the comment at `index.ts:296` and
  the framing in `tests/production-bundle.test.ts` to say "the executable
  bundle", and note in `.github/workflows/README.md` that the maps carry the
  full source on purpose.

Either way the comment needs to stop asserting something the deployed artifact
contradicts.

## Notes

Pre-existing on `wormeyman-space-age-support`; not introduced by any open PR.
Surfaced while reviewing #323, which is what prompted checking the maps at all.

byalex33:
The development-only Playwright API must be absent from executable production JavaScript, while public source maps intentionally retain the original source for browser debugging. The previous comment claimed the API never reached the production bundle, which was too broad: its source is present in the deployed maps.

Add a production build check using the website's real Vite configuration. It verifies seven stable test-API markers are absent from executable chunks, a production banner is present as a positive control, and linked source maps retain the original API source. The build stays in memory so running tests cannot overwrite a local `dist`. Document the public-map policy in the build config, startup comment, and CI guide.

Validation:

- `vp check --fix`: no format, lint, or type errors.
- `vp test`: 297 tests passed across 25 files.
- Replacing the DEV guard with `if (true)` fails the executable-code assertion on all seven markers; the positive control and source-map test still pass. Restoring the guard passes all three build tests.

This verifies compiled output, not production browser startup. Public source maps remain available intentionally; no confidentiality guarantee is made for their contents.

Closes #322. Closes #328. Addresses the artifact-publication portion of #352; local Worker compatibility and the scheduled cost check are separate follow-ups.

## Summary by CodeRabbit

* **Documentation**
  * Documented production bundle testing and source-map behavior.
  * Clarified development-only test hooks and test-file matching rules.

* **Tests**
  * Added coverage verifying production builds are valid and non-empty.
  * Confirmed test-only API markers are excluded from executable JavaScript while remaining available in intended source maps.


wormeyman (changes requested):
Reviewed `a51760e6`. This is a real test, not a shape of one: I confirmed the
negative control independently rather than taking the description for it. All
six markers survive minification because they are keys on the one object literal
that escapes to `window`, and I verified that empirically - five of them
(`viteStaticCopy` aside) appear verbatim in the committed 838 KB entry chunk. The
positive control is also stronger than the body claims: `Looking for the source?`
is top level in the *same module* as `testApi`, so it lands in the same chunk,
which means it proves the entry chunk was scanned rather than only that `dist`
was non-empty.

Two things to fix before merge, and one correction to the framing.

## 1. The test empties `packages/website/dist`, which is the deploy artifact

`tests/production-bundle.test.ts:76` calls `build({ root: websiteDir, logLevel:
'silent' })` with no `build.outDir`, so it resolves to Vite's default `dist` -
`packages/website/dist`. That is the same directory `npm run build:website`
writes and the same one `packages/worker/wrangler.jsonc:10` uploads
(`"directory": "../website/dist"`). Because the path is inside the config root,
`emptyOutDir` defaults to true, so the run **deletes** the existing tree first.
Nothing puts it back; `afterAll` only restores `NODE_ENV`. It is 163 MB here,
156 MB of that the `viteStaticCopy` of the exporter output.

Two local consequences:

- `npm run build:website`, then `npm run preview:website`, then `vp test` in a
  second terminal: the preview server's files vanish mid-run and requests 404
  or serve a truncated bundle. Nothing indicates a test did it.
- Interrupt `vp test` during the build and `dist` is left empty or half-written.
  `wrangler deploy` from `packages/worker` has no build step of its own, so a
  hand-run deploy would publish that tree as-is.

CI is safe and I checked rather than assumed - `checks` and `deploy` are
separate jobs on separate runners, and `.github/workflows/ci.yml` caches only
cargo and Playwright browsers, so nothing carries `dist` between them.

Fix is small: give the test its own output directory, e.g.
`build({ root: websiteDir, logLevel: 'silent', build: { outDir:
'dist-bundle-check', emptyOutDir: true } })`, point `assetsDir` at it, add it to
`.gitignore`, and `fs.rmSync` it in `afterAll`. `outDir` does not affect chunk
contents, so nothing is lost in fidelity to production.

## 2. Nothing proves a marker can ever match, so the six can rot into dead strings

`TEST_API_MARKERS` at `:53-61` is asserted only for *absence*, and
`git grep TEST_API_MARKERS` returns just its own two lines. There is no type
link either: `packages/website/src/index.ts:307` assigns through an `any` cast,
so the `testApi` literal is not checked against the `FbeTestApi` interface.

Renaming `spriteDataTally` means editing `index.ts`, `fbe-test-api.ts` and the
calling spec - all of which fail loudly. This file is not in that set and stays
green. Do that a few times as the API is reshaped, and every marker becomes a
string no build can contain; the guard then reverts, `leaked` is `[]`, and both
tests pass while the API ships.

Three lines close it: read `packages/website/src/index.ts` in `beforeAll` and
assert every marker appears in it. A rename is then a red `vp test` immediately.

## 3. The scope is narrower than the title

Worth correcting in the file comment rather than the code:

- A partial leak *within* `testApi` is structurally impossible - it is one
  object literal, no spreads, assigned once inside the single
  `import.meta.env.DEV`. So the six markers are six samples of one fact and the
  list never needs to grow. That is good, and worth saying, because the obvious
  future worry ("we added a 40th hook, update the list") is a non-issue.
- What the scan does *not* cover is a leak that is not that literal. Fourteen
  `testApi` keys delegate to `Editor` getters that exist only for specs, and
  class members on an instantiated class are not tree-shaken - `entityScreenPosition`,
  `exportEncodeCount`, `toolsPanelBounds` and four others are in the production
  bundle today. Unreachable without the module-local `editor` binding, so not a
  defect, but "does not carry the Playwright test API" means "does not carry the
  website-side glue object".
- The sourcemap is the third gap, and it is a live one. I filed it as #328 with
  the measurement rather than adding it here, since it predates this PR.

## Checked and clean

`NODE_ENV` handling is right and necessary: vitest sets it to `test`, Vite only
defaults it when unset, so without `beforeAll` the guard evaluates true and the
scan fails loudly rather than passing vacuously. `afterAll` runs inside the
runner's `finally`, so it restores on a thrown build too, and the default forks
pool with `isolate: true` means the mutation cannot reach a sibling file. The
`180_000` hook timeout is doing real work - the vitest default is 10s and this
would have been flaky without it. The refreshed comment in
`blueprint-files.test.ts` is accurate: 5 `.test.ts` under `tests/` now, 52
`.spec.ts`, and the `unit` include really is the mirror of
`testMatch: '**/*.spec.ts'`.

tests/production-bundle.test.ts
```diff
@@ -0,0 +1,121 @@
...
+            leaked,
+            `test-API symbols found in packages/website/dist/assets/*.js: ${leaked.join(', ')}. ` +
+                'The import.meta.env.DEV guard in packages/website/src/index.ts is not holding.'
+        ).toEqual([])
+    })
+
+    it('retains original source in public source maps for debugging', () => {
+        expect(sourceMapCount).toBeGreaterThan(0)
+        expect(bundledJs).toContain('sourceMappingURL=')
+        for (const marker of TEST_API_MARKERS) expect(mappedSource).toContain(marker)
```
wormeyman, line 148: This closes most of the gap. But `mappedSource` is `sourcesContent` from every
chunk, so it covers the editor package's sources too, not only
`packages/website/src/index.ts`. Two markers are mentioned in the editor as well:

- `spriteDataTally` in `packages/editor/src/containers/EntitySprite.ts`
- `overlayInfoTally` in `packages/editor/src/containers/EntityContainer.ts`

Those two stay "present" after they are renamed in `index.ts`, which is the rot
this check exists to catch. Measured: renaming `spriteDataTally` in `index.ts`
only, leaving `TEST_API_MARKERS` untouched, leaves all three tests green.

Reading the one file closes all seven:

```ts
const indexSource = await fs.promises.readFile(
    path.join(websiteDir, 'src/index.ts'),
    'utf8'
)
for (const marker of TEST_API_MARKERS) expect(indexSource).toContain(marker)
```

Worth its own `it` rather than living here. As written the marker check rides on
the source-map assertion, so if public maps are ever turned off - a fair response
to #328 - `mappedSource` goes empty, this test fails, and the marker guard goes
with it.

tests/production-bundle.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        // nothing in it to match.
+        expect(bundledJs).toContain('Looking for the source?')
+        expect(bundledJs.length).toBeGreaterThan(100_000)
+    })
+
+    it('does not carry the Playwright test API in executable JavaScript', () => {
+        const leaked = TEST_API_MARKERS.filter(marker => bundledJs.includes(marker))
+        expect(
+            leaked,
+            `test-API symbols found in packages/website/dist/assets/*.js: ${leaked.join(', ')}. ` +
```
wormeyman: Small one from the `write: false` change: this message points at
`packages/website/dist/assets/*.js`, which the test no longer writes or reads.
Anyone debugging a failure will go and look at a directory holding an unrelated
older build. Naming the in-memory chunks matches what was actually scanned.

The comment at :29-31 drifted the same way. It says the positive control guards
against "a grep over an empty or partial `dist`", and there is no dist in the
picture now.

wormeyman (commented):
Re-reviewed `918514db`. Two of the three items are closed, and I checked them by
running the test rather than reading the commit subjects.

Finding 1 is fixed, and `write: false` is better than the separate `outDir` I
suggested, because nothing is written at all. Verified against a real 163 MB
`packages/website/dist`: 3114 files before, 3114 after, asset mtimes unchanged.

The build behind it is genuine, which I checked because 0.75 s for the whole run
looked too fast: 344 ms, 13 chunks, 1,105,409 bytes of code, an 839,761-byte entry
chunk, PixiJS present, 12 maps. The negative control works too - replacing the
guard with `if (true)` fails the scan and names all seven markers.

Finding 3 is addressed in the header, the test name and the workflow README.

Finding 2 is 5 of 7 closed, and the two still open are the ones the finding was
about. Details inline.

tests/production-bundle.test.ts
wormeyman: Pushed this onto your branch as 236fc68a (maintainerCanModify was on), so #322 can close without another round trip. It is the `it` sketched above, reading `packages/website/src/index.ts` directly, with one change from the sketch: it matches whole words. A plain `includes` let `spriteDataTally` renamed to `spriteDataTallyRenamed` pass as a substring hit, which is the same hole the original `toContain` had.

Checked both ways. Renaming `spriteDataTally` in index.ts alone now fails only the new test, naming that marker. Replacing `import.meta.env.DEV` with `true` fails only the leak test, listing all seven. `vp check` is clean and `vp test` is 344 passing.

tests/production-bundle.test.ts
wormeyman: Done in the same commit (236fc68a). The message now says "the production build's JavaScript chunks", and the two comments that said dist say "an empty or half-built set of chunks" instead.

tests/production-bundle.test.ts
```diff
@@ -0,0 +1,150 @@
...
+            missing,
+            `markers not found in ${indexPath}: ${missing.join(', ')}. ` +
+                'Either the test API renamed them or the guard above is scanning for the wrong names.'
+        ).toEqual([])
+    })
+
+    it('retains original source in public source maps for debugging', () => {
+        expect(sourceMapCount).toBeGreaterThan(0)
+        expect(bundledJs).toContain('sourceMappingURL=')
+        for (const marker of TEST_API_MARKERS) expect(mappedSource).toContain(marker)
```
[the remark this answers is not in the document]
wormeyman: Skipping this one. The test does not claim a per-chunk source-map contract. It claims that the executable JavaScript is free of the test API and that the public maps still carry the original source for debugging (#328), which the three aggregate checks cover.

Measured on this head with `vp build`: 14 JavaScript chunks, 12 `.js.map` files, and every chunk that carries a `sourceMappingURL` resolves to an emitted map. The two without one are `transcoder.1.16.4-*.js` (the prebuilt basis transcoder) and `webworkerAll-*.js` (PixiJS's worker bundle). The per-chunk check proposed here would fail on both today, so it would be a new assertion about third-party worker files rather than a tighter version of this test. If that contract is wanted it belongs in its own issue, not in a contributor's PR about the DEV guard.

PR #325, opened 2026-09-03
Make ExportDialog's re-encode debounce a frame-checked deadline, and stop the burst spec racing it

### LINKED ISSUE #313: quick-actions.spec.ts:673 races the 500 ms debounce it is measuring
A flaky assertion with a known mechanism, not a mystery. The spec's own header already describes this failure mode for a *sibling* check that was deleted because of it; this one was left in and can fail the same way.

## The failure

`tests/quick-actions.spec.ts:611:5 - a burst of edits inside one debounce window collapses to a single re-encode (#242 review follow-up)`

```
Error: expect(received).toBe(expected) // Object.is equality
Expected: 2
Received: 3

  672 |     await page.waitForTimeout(700)
> 673 |     expect(await page.evaluate(() => window.__fbe_test.exportEncodeCount())).toBe(2)
```

Seen on Playwright shard 3/4, run 33589280819. The PR it fired on changed only `packages/worker/wrangler.jsonc`, a Cloudflare config file that neither the editor nor the dev server loads, so there was no mechanism by which it could be related. **Re-running the failed job with zero code change passed all four shards.**

## The mechanism

The test depends on a wall-clock wait landing inside a 500 ms window:

1. Open the export dialog. Count is 1.
2. Delete an entity. The 500 ms debounce timer starts.
3. `await page.waitForTimeout(200)` - intended to sit inside that window.
4. Press undo, which should restart the debounce so both edits coalesce.
5. Poll until the count is 2, then wait 700 ms and assert it is still 2.

`page.waitForTimeout(200)` has no upper bound on how much *more* than 200 ms elapses before the following step runs. Under shard CPU contention it can exceed 500 ms, and then the delete's debounce has already fired on its own. The count reaches 2 from the delete, so the poll at line 665 passes for the wrong reason, and the undo's separate re-encode lands during the 700 ms wait, making 3.

That is exactly what the spec header at lines 633-647 already says about the check that used to sit between steps 3 and 4:

> That raced the same 500 ms debounce it was trying to observe: `page.waitForTimeout` is a wall-clock wait with no floor on how much *more* than 200 ms elapses before the following `page.evaluate` round-trip actually runs, and under enough system load (measured reproducibly as part of a full, 54-test shard, never in isolation) that slack alone was enough to cross 500 ms.

The sibling was removed. The 200 ms wait it raced is still there, and lines 665 and 673 both still depend on it.

## What not to do

Do not add a retry or lengthen the timeout. Per this repo's own rule, an intermittent spec usually has a deterministic bug under it, and here the mechanism is already identified: the test cannot control whether step 3 stays inside the debounce window.

The constraint a fix has to satisfy is that the *coalescing* is what gets asserted, without the assertion depending on a wall-clock wait beating a timer. Whether that means making the debounce window injectable, exposing when it fires, or asserting something other than an exact count, is a design call for whoever owns this spec - I have not proposed a patch.

## Rate

Unknown. I saw it once, and one no-change re-run passed. The header's note that the sibling reproduced "as part of a full, 54-test shard, never in isolation" suggests it needs the loaded-shard conditions rather than being reproducible on demand.

https://claude.ai/code/session_014iKcwmj7ZZ1sZJyDtb7VK8

byalex33:
Closes #313.

The burst-edit regression test raced the 500 ms export debounce on loaded CI shards. ExportDialog now uses one `performance.now()` deadline, extended whenever its existing frame poll observes a history revision change. A pending encode can be inspected and flushed through the test API.

The regression test widens the debounce beyond its test timeout, deletes a chest, undoes, then redoes. It waits for each entity change and asserts that undo and redo strictly advance the pending deadline, with no intervening encode. One flush must replace the original two-chest export with a one-chest export; a second flush must do nothing. This detects both stale output and a deadline that only arms when idle, without a fixed sleep or treating a stable history revision as proof that frames settled.

The page-global debounce override is restored in `finally` and rejects negative/non-finite durations. Pending/flush hooks return `undefined` when no dialog is open. The unused history-revision hook was removed.

The production debounce remains 500 ms, but expiry is checked on animation frames. Hidden tabs may defer export refresh until rendering resumes; foreground scheduling follows the application's ticker, and closing the dialog removes its listener.

Validation:
- All 15 quick-actions Playwright tests passed; the burst regression also passed five consecutive runs.
- Both review mutations failed at their intended assertions: incrementing the encode count without refreshing text, and extending the deadline only when idle.
- All 216 unit tests passed with the oracle fixture checked out using its committed LF line endings (Windows CRLF checkout otherwise causes its existing byte-comparison failure).
- Full repository formatting, lint, and type checks passed after normalizing local checkout line endings to LF (no committed content change). Local checks used the existing installed dependencies; GitHub CI validates the lockfile installation.

## Summary by CodeRabbit

- **New Features**
  - Added visibility into when pending export re-encoding is scheduled.
  - Added controls to check, flush, and configure export re-encoding debounce behavior.
  - Pending export controls now indicate when no export dialog is open.
  - Added desktop controls for opening book details and blueprint information.

- **Bug Fixes**
  - Improved export re-encoding during rapid edits, deletions, undo, and redo actions.
  - Prevented outdated export text updates after newer changes or dialog closure.
  - Consolidated successive edits into a single re-encoding operation.

- **Tests**
  - Added validation for invalid debounce durations and coverage for timing and final export output.


wormeyman (changes requested):
Reviewed `87bb28cc`. The debounce redesign is sound and I traced it rather than
reading the description for it: the ticker listener is removed on `destroyed`,
pixi emits that *before* `removeAllListeners()` so the `once` handler always
runs, and every close path funnels through `Dialog.close()`. `performance.now()`
is the right clock. Nothing is captured at arm time, so a fire cannot export
stale state. And there is no re-encode loop, because `generateIcons` on this base
writes a plain `Map` with no `history.` call, so an encode cannot move the
revision the poll reads.

The diagnosis in #313 is right too, and this fixes the mechanism rather than
hiding the race: making the deadline flushable takes the timer off the test's
critical path entirely, instead of widening a tolerance.

Two problems in the spec, which is the deliverable here.

## 1. The closing assertion cannot fail

`tests/quick-actions.spec.ts:686-695`.

The burst is delete then undo, which returns the blueprint to exactly the state
encoded at open. With the window widened to 120 s nothing re-encodes in between,
so `m_TextInput.text` holds the open-time string - a 2-chest blueprint - for the
whole test. The post-flush re-encode also produces a 2-chest blueprint. The
assertion is satisfied by the stale text and the fresh text alike.

Mutation that stays green: replace the `refreshText` call in
`flushPendingReencode` with `this.m_EncodeCount += 1`. Lines 680, 681, 682 and
683 all pass, and 686-695 decodes the stale open-time string to 2 entities.
Nothing in this test observes the flush's *output*, only its counter.

Ending the burst somewhere other than the start state fixes it - delete, undo,
redo leaves 1 entity against a stale 2, and keeps the "one keypress, no pointer
targeting" property the comment at `:627-628` is protecting.

Related, minor: the `catch { return -1 }` at `:691` turns a decode failure into
a poll that runs to the 60 s expect timeout and then reports `-1` rather than the
parse error.

## 2. Nothing tests the property the redesign is built on

`ExportDialog.ts:152-153` says the debounce is "one due-time that every further
change pushes forward", and `:674-675` says "the undo pushed the same debounce
out". Neither is observable: `reencodePending` is `m_ReencodeDueAt !== undefined`
on a single scalar, and the due *time* is never exposed, so both assertions hold
whether or not the deadline moved.

Mutation that stays green here and in the sibling test at `:561-612`: arm only
when idle.

```ts
if (this.m_ReencodeDueAt === undefined)
    this.m_ReencodeDueAt = performance.now() + reencodeDebounceMs
```

Under a 120 s window the deadline is far out either way, pending is true, the
count is 1, and one flush gives one encode. In production it means a 2-second
drag re-encodes every 500 ms *mid-drag* on partial state - exactly the regression
the debounce exists to prevent.

The old test had the same gap, so this is not something the PR opened. But it
asserts the property more confidently than the code did before, and the repo has
already had a commit for comments that outrun their code (`74de74a1`, #284).
Exposing the due time or an armed count, and asserting it strictly increased
across the undo, pins it with no timing involved.

## Smaller things

- `setExportReencodeDebounceMs` never restores. `reencodeDebounceMs` is a
  module-level `let`, and `UIContainer.setExportReencodeDebounceMs` is the one
  hook of the three that does not go through `this.exportDialog?`, so the
  override outlives the dialog and every later instance. Safe today *only*
  because each Playwright test gets a fresh page - a `describe.serial` block
  sharing one page, or a test appended after this one, would inherit the 120 s
  window and the sibling at `:605` would fail as "the debounce never fired". A
  clause in the `fbe-test-api.ts` doc saying the override is page-global, or an
  `afterEach` restoring 500, removes the landmine.
- `exportReencodePending` and `flushExportReencode` return `boolean`, collapsing
  "no dialog open" into `false`. The sibling `exportEncodeCount` deliberately
  returns `number | undefined` and its comment says why. As written,
  `flushExportReencode() === false` cannot tell "nothing armed" from "the dialog
  is gone", which matters because `Editor.loadBlueprint` calls `Dialog.closeAll()`.
- `historyRevision`'s doc in `Editor.ts:392` and `fbe-test-api.ts:80` says
  polling it to a fixed value shows "an edit and its follow-up frames have
  settled". `History.revision`'s own comment at `History.ts:218-235`, written to
  stop this exact over-claim, says it supports only "is this different from what
  I last saw" - it does not move on an empty outermost commit, and crossing
  `MAX_HISTORY_LENGTH` moves it backward. The test is not wrong, because a
  premature conclusion fails loudly on the next line rather than passing; the
  comment is.
- `revisionSettled` is called after the delete with no landed-check, but after
  the undo with one (`entityScreenPosition(1) !== undefined`). The asymmetry is
  invisible today because `exitDeleteMode` removes entities synchronously before
  `mouse.up()` resolves, so the revision has already moved. A matching poll after
  the delete costs nothing and makes the soundness local rather than dependent on
  that.
- The PR body says behaviour is "otherwise identical". The deadline now matures
  only on a ticker tick, and pixi's ticker is rAF-driven, which browsers stop in
  a hidden tab - so a backgrounded tab defers the re-encode instead of running
  it. Probably an improvement, since it keeps a ~300 ms serialize off the main
  thread of a tab nobody is looking at, but it is a change. The parenthetical
  "while the dialog - and so the ticker - is up" also reads backwards: the ticker
  runs because `Application` holds a render listener, not because this dialog is
  open.
- `setExportReencodeDebounceMs(NaN)` wedges the dialog permanently - `performance.now()
  + NaN` is `NaN`, the `>=` is always false, so `reencodePending` stays true
  forever. Test-only input and no caller passes one, so it is bottom of the list.

## Not blocking, for the record

The coverage claim I expected to have to make does not apply: the ~500 ms window
maturing on its own is still exercised, by the untouched sibling at `:600-607`
which polls `exportEncodeCount` to 2 with no override. What is lost is the
"settles at 2 rather than passing through it" tail, and nothing anywhere pins the
500 ms duration itself - which was true before this PR too.

wormeyman:
@coderabbitai review

wormeyman:
Following up because this has gone quiet for a couple of days. The review from
2026-09-04 is still the open item - the head here, `87bb28cc`, predates it, so
nothing has landed against it yet.

CodeRabbit approved this on 2026-09-05 at my request. That approval does not cover
either of the two problems I raised, both of which are about what the spec proves
rather than the code under it. I traced the debounce redesign itself and it is
sound.

The two asks:

1. End the burst somewhere other than the state it started in. Delete, undo, redo
   leaves 1 entity against a stale 2, so the closing assertion can tell the flushed
   text from the open-time text. As written both are a 2-chest blueprint and the
   assertion holds either way.

2. Expose the due time, or a count of how many times the deadline was armed, and
   assert it strictly increased across the undo. Right now `reencodePending` is
   only `m_ReencodeDueAt !== undefined`, so "the undo pushed the same debounce out"
   is not observable, and arming only when idle stays green.

byalex33:
@coderabbitai review

byalex33:
@coderabbitai review

PR #326, opened 2026-09-03
Count unique visitors from both sides of the JavaScript boundary

wormeyman:
Nothing measured visitors. Workers metrics count requests — the figures `wrangler.jsonc` quotes — and requests are not people: 9,097 asset files and a crawler spike move that number with no visitor behind it.

Two counters, deliberately disagreeing. The Worker counts requests that never ran JavaScript; the beacon counts sessions that did. Neither is exact, and the gap between them is the useful part.

## The beacon

`packages/website/vite.config.js` injects Cloudflare Web Analytics at build time when `CF_BEACON_TOKEN` is set, which only the deploy job does. Unset, the tag is omitted and the build is otherwise identical — so a fork stays green and a developer's page loads never reach the production site's analytics.

The CSP in `public/_headers` already allowed both hosts this needs, from before this change. That is exactly the arrangement that breaks quietly: tightening it would fail no build, fail no Playwright run, and stop analytics reaching Cloudflare with no error anywhere but a visitor's console. `tests/visitor-count.test.ts` pins the tag and the header together.

## The Worker count

One GET of `/` per IP + User-Agent + Accept-Language per UTC day, deduped in the Cache API and written to the `fbe_unique_visitors` Analytics Engine dataset.

Three choices worth keeping:

- **Dedupe at write time, not in the query.** Analytics Engine has no `uniq()` and no `COUNT(DISTINCT)`, so a data point has to already mean one visitor-day; reading it is then `SUM(_sample_interval)`.
- **Cache API, not KV, for the dedupe store.** KV would dedupe globally where the cache is per-colo, but a KV binding needs an id pasted in from the dashboard before a deploy works. A count that runs a few percent high beats a deploy of the whole site failing on a missing setup step. An Analytics Engine dataset needs no id at all — its first write creates it.
- **No visitor identifier is stored.** The fingerprint is a cache key inside its 24h window and never reaches the dataset, which carries the day, the country and the number 1. No cookie, so no consent banner.

Counting runs in `ctx.waitUntil` with the whole body wrapped, so it is off the response path and cannot fail a page load. `isPageView`, the fingerprint and the cache key live in a pure module for the reason `proxyTarget.ts` does: `packages/worker` is linted and collected by nothing, so a pure module is the only part a test can reach.

## Known inaccuracies, all documented in the source

- Per-colo dedupe over-counts a visitor who crosses colos in one day. One direction only, and normally one colo per person.
- The Cache API is inert on `workers.dev`, so a versioned preview hostname counts every request rather than once a day. Production is a custom domain, so this affects preview traffic only.
- `cache.match` and `cache.put` are two calls with nothing holding the gap between them, so a prefetch racing the navigation that follows it can write one visitor-day twice. Same direction as the other two; the window is the few milliseconds between the calls.

## Verification

- `vp check .` clean: 235 files formatted, 0 warnings/lint/type errors in 192 files.
- 18 new unit tests pass (233 → 234 across the suite); `tsc --noEmit` at 0 for both the worker and editor projects.
- The two source-reading tests are mutation-checked: writing `max-age` as a literal, dropping the `cache-control` header, and redeclaring `DEDUPE_WINDOW_SECONDS` locally each turn them red.
- A build with `CF_BEACON_TOKEN` set emits the beacon tag; one without emits nothing. Both checked in the built `dist/index.html`.
- The two failures in `scripts/localpreview.test.mjs` are the test container having no IPv6 (`EAFNOSUPPORT ::1`) and fail identically at the branch point.

Not verifiable from here: the Cache API lookup, the `writeDataPoint` call and `ctx.waitUntil` run nowhere but production — the same gap `tests/corsproxy.test.ts` records for the proxy.

## After merge, or nothing reports

1. ~~Create the Web Analytics site in the Cloudflare dashboard and add its token as the `CF_BEACON_TOKEN` repo secret.~~ Done — the secret exists, so the deploy that follows this merge injects the beacon.
2. Analytics Engine has no dashboard view. Read that number over the SQL API with the query in the `recordVisit` comment in `packages/worker/src/index.ts`, or through Cloudflare's Grafana plugin. The zone's Analytics & Logs → Traffic tab already shows IP-based unique visitors retroactively, with no deploy needed.

https://claude.ai/code/session_01Kftg4y5VSxEvkFFez2KZbs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -307,8 +307,15 @@ jobs:
       - name: Install dependencies
         run: vp install --frozen-lockfile
 
+      # CF_BEACON_TOKEN injects Cloudflare Web Analytics' beacon into
+      # index.html (packages/website/vite.config.js). Deploy-only on purpose:
+      # the `checks` job builds nothing that ships, and a PR build must not
+      # report page loads as though they were the live site. An unset secret
+      # omits the tag and the build is unchanged, so this stays green in a fork.
       - name: Build website
         run: npm run build:website
+        env:
+          CF_BEACON_TOKEN: ${{ secrets.CF_BEACON_TOKEN }}
 
       - name: Deploy worker to Cloudflare
         uses: cloudflare/wrangler-action@ebbaa1584979971c8614a24965b4405ff95890e0 # v4
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -230,6 +230,33 @@ is `packages/worker/wrangler.jsonc`; GitHub Actions deploys after checks and e2e
 tests pass. Required secrets are `CLOUDFLARE_API_TOKEN` and
 `CLOUDFLARE_ACCOUNT_ID`.
 
+## Visitor counts
+
+Two counters measure the same thing from opposite sides, and they are meant to
+disagree.
+
+- Cloudflare Web Analytics. `packages/website/vite.config.js` injects the beacon
+  at build time when `CF_BEACON_TOKEN` is set, which only the deploy job does.
+  Unset, the tag is omitted and the build is otherwise identical. Read it in the
+  Cloudflare dashboard. It counts sessions that ran JavaScript, so ad blockers
+  and crawlers are missing from it.
+- A server-side count in the Worker. `packages/worker/src/visitorCount.ts` holds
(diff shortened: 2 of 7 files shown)


wormeyman:
@claude review

wormeyman:
@claude review

Issue #329, opened 2026-09-04
Triage of all 38 open issues left behind by unmaintained upstream, plus four bugs it surfaced in our own fork


wormeyman:
Upstream `teoxoy/factorio-blueprint-editor` was marked unmaintained on 2026-08-16 (`62d8b92`, upstream #276) with **38 issues left open**. This is a full triage of all 38 against our fork, plus a deep read of upstream #277 (the doorbell.io feedback aggregator). Umbrella for #247.

**Method.** Five parallel agents read every upstream issue, then verified each claim against our source rather than against the issue text. Every "already done" / "obsolete" verdict below carries a `path:line` citation. Nothing was run in a browser — no `node_modules`, no Factorio data server, no Factorio install — so verdicts are source-and-fixture reads. The four blueprint strings below were each verified locally through `zlib.inflateSync` + `JSON.parse`.

**Headline:** the most valuable output is not a port list. It is **four latent bugs in our own fork** that the upstream reports led to, one of which is confirmed and untested (§4).

---

## 1. Recommended order

| | Item | Why first | Effort |
|---|---|---|---|
| 1 | **Underground belts never pair horizontally** (§4.1) | Confirmed live bug, no test coverage, explains 3 separate upstream reports | S |
| 2 | **Firefox Ctrl+V does nothing** (§4.2) | Confirmed, one-line class of fix, affects every Firefox user | S |
| 3 | **`serialize()` silently drops `parameters`** (§4.3) | Silent data loss on export | S |
| 4 | Beacon supply area is drawn and counted wrong (upstream #263) | Two call sites, one shared number, both wrong | S |
| 5 | Duplicate logistic filter slots (upstream #242) | Our own oracle fixture already states the game's rule | S |
| 6 | Inserter stack size editor (upstream #45) | 1007 corpus entities carry the field; getter exists, no setter | S |
| 7 | Hi-res / transparent image export (upstream #228) | `resolution: 1` is hardcoded; backgrounds are children of the captured container | S |
| 8 | Train stop station name rendering (upstream #19) | The label renderer already exists in our fork | S |
| 9 | Pasted electric poles do not connect (upstream #230) | One flag, but needs a duplicate-wire guard | S–M |
| 10 | Entity count on drag-select + blueprint BOM (upstream #246) | Best value-per-effort feature in the set | S–M |

---

## 2. Already done or fixed in our fork — close upstream as answered

| Upstream | Verdict | Evidence |
[...]

PR #333, opened 2026-09-04
Put the beacon supply area on the beacon, at its real 9x9

wormeyman:
The beacon's supply area is drawn one tile too small and counted half a tile off centre. This is upstream issue [teoxoy/factorio-blueprint-editor#263](https://github.com/teoxoy/factorio-blueprint-editor/issues/263), and it reproduces here.

Part of #329 (§4.4).

## Ground truth

From `data.json`: the beacon has `supply_area_distance = 3` and `collision_box = [[-1.2,-1.2],[1.2,1.2]]`, so a 3x3 entity with a **9x9** supply area. Neither 2.0 nor 2.1 changed it.

## The triage's premise was half right

It said "electric poles are correct under the same formula because they are 1x1 or 2x2". They are not correct under the same formula - **they are correct because the pole branch never used the beacon's formula**. `supply_area_distance` means two different things on the two prototypes, and typed-factorio quotes the docs verbatim for the pole:

> `ElectricPolePrototype.supply_area_distance` - "Corresponds to **half** of the 'supply area' in the item tooltip. If this is 3.5, the pole will have a 7x7 supply area."

Entity size plays no part there. On `BeaconPrototype` the distance is measured outward from the footprint, so size does count.

Applying the suggested `supply_area_distance + tile_size/2` to poles would have broken all four: small pole `2.5 + 0.5` = 6x6 where the game gives 5x5, big pole `2 + 1` = 6x6 where the game gives 4x4. The proposed formula is right **for the beacon only**.

## The formula, and how it was checked

- Beacon: `halfExtent = supply_area_distance + max(width, height) / 2` → `3 + 1.5` = **4.5**, a 9x9 area.
- Pole: `halfExtent = supply_area_distance`, unchanged.

A supply area lines up with the tile grid, so `centre ± halfExtent` has to be a whole number. A 1x1 pole sits at `x.5` with distances 2.5/3.5; a 2x2 pole sits on a corner with 2/9; a 3x3 beacon sits at `x.5` and gets 4.5; a hypothetical 2x2 beacon sits on a corner and gets `3 + 1` = 4. All land on the grid. Swap the two rules and every one lands on `x.5`. That parity check is in the test suite, not just in this description.

Verified against `data.json`: beacon 3 / 3x3; small pole 2.5 / 1x1; medium 3.5 / 1x1; big 2 / 2x2; substation 9 / 2x2 - giving 5x5, 7x7, 4x4 and 18x18, matching the game's published figures. Those five are the only entities carrying the field.

## What changed

`packages/editor/src/core/supplyArea.ts` (new) is the one place the number now lives - `supplyAreaHalfExtent`, `supplyAreaBounds`, `suppliesEntity`. Pure, no pixi, no globals, following `throughput.ts` and `zoomLevels.ts`. Overloads return a plain `number` inside an `isBeacon`/`isElectricPole` guard so neither call site needs a fallback.

- **`UnderlayContainer.ts`** - both branches take their radius from the helper. Pole values are unchanged; the beacon goes 4 → 4.5, so the drawn square goes 8x8 → 9x9.
- **`EntityInfoPanel.ts`** - `findNearbyBeacons` is now one `suppliesEntity` call. Two further corrections beyond the geometry: it read `FD.entities.beacon` rather than the beacon in hand, and it tested four corners for containment instead of testing footprints for overlap.

That second one matters. **Fixing the rectangle alone would have introduced a new bug**: `Rectangle.contains` treats its left and top edges as inclusive, so a neighbour merely touching them would have been counted.

Precise old-vs-new for the panel: the shipped aura was `[cx-2.5, cx+6.5)` for a beacon at `cx=1.5`. West and north came out right by luck - half a tile of slack is less than a tile. East and south were **one tile too far**, so an entity a full tile outside the beacon's reach was credited with its modules.

## Tests

`packages/editor/src/core/supplyArea.test.ts`, 14 cases built from the real `data.json` numbers: beacon half-extent and an even-sized synthetic beacon, all four poles asserted at 5x5 / 7x7 / 4x4 / 18x18, actual tile spans rather than just "returns something", the whole-tile grid parity check across all five real suppliers, `suppliesEntity` boundaries on all four sides (last tile in true, one past false), a 3x3 assembler in the aura corner, and a 1x20 entity crossing the aura with all four corners outside.

Reverting the beacon formula to the old `+ 1` turns 4 of the 14 red.

```
$ vp check
pass: All 237 files are correctly formatted
pass: Found no warnings, lint errors, or type errors in 194 files

$ vp test
 Test Files  21 passed (21)
      Tests  248 passed (248)

$ npx tsc --noEmit -p packages/editor/tsconfig.json
0 errors
```

## Fixtures

**Nothing pinned moves.** `supply_area_distance` occurs in only those two source files. The five beacon mentions in `tests/__fixtures__/sprite-data.json` are sprite layers. `core/generators/beacon.ts` hardcodes its own 3x3 logic and never reads the field, so `generators/__fixtures__/expected-output.json` is unaffected. No spec references `UnderlayContainer` or `findNearbyBeacons`.

## Not verified

The 9x9 underlay square is checked by arithmetic here, never by eye - Playwright cannot run on the machine this was built on. Per CLAUDE.md's "drive the editor and print numbers" rule, **worth one look at a beacon on the deploy preview before merging.**

### DIFF
--- a/packages/editor/src/UI/EntityInfoPanel.ts
+++ b/packages/editor/src/UI/EntityInfoPanel.ts
@@ -1,4 +1,4 @@
-import { Container, Rectangle, Text } from 'pixi.js'
+import { Container, Text } from 'pixi.js'
 import FD, {
     getModule,
     isBeacon,
@@ -15,6 +15,7 @@ import { Entity } from '../core/Entity'
 import F from './controls/functions'
 import { Panel } from './controls/Panel'
 import { styles } from './style'
+import { suppliesEntity } from '../core/supplyArea'
 import { beltThroughput, inserterThroughput } from '../core/throughput'
 
 /*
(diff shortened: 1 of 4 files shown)


wormeyman:
@claude Review; check out https://wiki.factorio.com/Beacon and https://lua-api.factorio.com/latest/prototypes/BeaconPrototype.html as a comparison.

PR #332, opened 2026-09-04
Add four reporters' blueprints from upstream #277 to the corpus

wormeyman:
Adds four of the seven blueprints that upstream issue #277 collected under "Blueprints that don't load" to the committed corpus, and re-records the five pinned fixtures that the larger corpus moves.

Part of #329 (§8).

## Where these came from

Upstream `teoxoy/factorio-blueprint-editor` was marked unmaintained on 2026-08-16. Issue #277 is a dump of the feedback its author had collected privately, and one section of it is seven blueprint strings from users reporting that the editor would not load them.

**They all load here.** Every prototype name across all 4,046 entities is in our `data.json` except `logistic-chest-storage`, which `nameMigrations.ts` already renames to `storage-chest` under its pre-2.0 gate. Upstream's failures were a 1.1-era editor being fed 2.0 blueprints, not a defect that reproduces in this fork. They are wanted as corpus fixtures, not as bug reports.

Every string was re-verified locally through `zlib.inflateSync` + `JSON.parse` before being committed. The recovery went through a summarising web fetch on the first attempt and three of four strings came back silently corrupted, so nothing here is trusted that did not round-trip.

## What was added, and why each one earns its place

New collection folder `test-blueprints/UPSTREAM-277/`. The folder is mandatory: `discoverBlueprintFiles` walks exactly one level down and now throws on a top-level `.txt` (#190), so a stray file fails CI rather than being skipped.

| File | Version | Entities | Why |
|---|---|---|---|
| `combat-robot-capsules-1-1-69.txt` | 1.1.69 | 186 | The first real blueprint reaching **any** pre-2.0 branch. Carries 4x `logistic-chest-storage` and 8 `control_behavior.filters` with no `sections` |
| `corner-defense-1-1-34.txt` | 1.1.34 | 438 | 24 gates, against 240 in the entire existing corpus, plus 196 `stone-wall`. Gate base selection reads the grid |
| `beacon-mall-2-0-43.txt` | 2.0.43 | 1,145 | 230 undergrounds against 440 belts - 52%, where the corpus average is 26% |
| `mining-outpost-2-0-73.txt` | 2.0.73 | 1,368 | Contributes the only entity name new to the corpus across all seven: `electric-mining-drill`, 175 of them, previously synthetic-only |

**Three were rejected rather than added.** A bloated corpus costs time on every run:

[...]

Issue #288, opened 2026-08-31
The unguarded-icon throw is in five more places than #286 names, including the branch #285 did not reach


wormeyman:
Found by writing the test for #286 and watching it still fail. All five are
measured, each with the stack that produced it.

#286 says `Recipe.ts:38` is "the one place that still has it". It is one of six.

## The four more icon slots

Same shape as #286 exactly - a name out of the blueprint reaching a bare
`F.CreateIcon`, on a path with no `try` above it, so the cost is the whole
dialog rather than one icon.

| site | reached by | blueprint field |
| --- | --- | --- |
| `DisplayPanelIcon.ts:39` | `DisplayPanelEditor` | `icon` |
| `Modules.ts:40` and `:68` | `Editor.addModules` | `items` |
| `Filters.ts:242` | `ChestEditor` | `request_filters` |
| `Filters.ts:251` | `SplitterEditor`, `InserterEditor` | `filter` |

**The display panel one is the half PR #285 did not reach, and it is the
ordinary case.** `DisplayPanelEditor` has two branches. `if (connected)` draws
the read-only conditions list and holds the two sites #285 guarded; that branch
`return`s, and an **unconnected** panel falls past it to
`new DisplayPanelIcon(entity)` at `:141`, whose `F.CreateIcon` is bare. Both
display panel specs on record set `connect_to_logistic_network: true`, so both
take the guarded branch and neither can see it. Measured: a panel with
`icon: { type: 'space-location', name: 'nauvis' }` and no logistic connection
throws `No item, fluid, recipe, signal or inventory group named nauvis` and
opens no dialog - which is #280's bug, still live, in the editor #280 was about.

Note `Entity.displayPanelIcon` reads `m_rawEntity.icon || control_behavior
.parameters?.[0]?.icon`, so this slot can also be fed by the parameters a
connected panel uses.

## The two that are not icon slots

These are why guarding `Recipe.ts` alone does **not** make the machine editor
open, and neither was visible from reading the issue.

**`Preview.generatePreview`** calls the *static*
`OverlayContainer.createEntityInfo` with nothing around it, while
`OverlayContainer`'s own instance method wraps the identical call in try/catch
at `OverlayContainer.ts:493`. The overlay draws recipe, module and filter icons,
so it throws on the same names. `Preview` is built by `Editor`'s base
constructor at `Editor.ts:25`, so this one costs **every** editor its dialog:

```
CreateIcon (functions.ts:148)
[...]

Issue #286, opened 2026-08-30
Recipe.ts throws on an unknown recipe name, costing the whole machine editor


wormeyman:
The same bug class #280 fixed for `DisplayPanelEditor`, in the one place that
still has it.

`packages/editor/src/UI/editors/components/Recipe.ts:38`:

```ts
this.content = F.CreateIcon(recipe)
```

`F.CreateIcon` ends in a bare `throw` (`packages/editor/src/UI/controls/functions.ts:229`)
when the name is in none of `FD.items`, `FD.fluids`, `FD.recipes`, `FD.signals`
or `FD.inventoryLayout`. The recipe name comes from the blueprint.

`F.SafeIcon` now exists in `functions.ts` for exactly this - it warns naming the
icon and returns an empty container. `DisplayPanelEditor` uses it; this does not.

## Two call sites, and the second is the nastier one

```
:17   this.updateContent(this.m_Entity.recipe)              // constructor
:20   this.onEntityChange('recipe', recipe => this.updateContent(recipe))
```

- **From the constructor.** `MachineEditor` and `TempEditor` call `addRecipe()`
  in their own constructor bodies, so an unknown recipe means the editor throws
  on the way up and the user gets **no dialog at all** from clicking the machine.
  `UIContainer.createEditor` is a bare call with no `try` above it, on purpose
  (see the comment there), so the click just does nothing.
- **From the entity-change handler.** A recipe changing to an unknown name
  throws inside an event listener, on a dialog that is already open. Nothing
  catches that either.

## Not a registry leak any more

Worth stating so nobody re-files this as #280. Since `3785461d` a dialog
registers on the pixi `added` event rather than in its constructor, so a
constructor that throws no longer leaves a phantom in `Dialog.s_openDialogs`.
What is left is the narrow half: the throw still costs the whole editor where it
should cost one icon.

## Fix

`F.SafeIcon(recipe, () => F.CreateIcon(recipe))` at `:38`, matching what
`DisplayPanelEditor` does.

## Coverage

`tests/dialog-registry-leak.spec.ts` already builds an assembling machine
carrying a recipe name FD does not have - it is the fixture for that spec's
general half, precisely because this throw is still reachable. A test for this
fix has its fixture waiting.

PR #334, opened 2026-09-04
Read a book's active_index as an inventory slot, and stop losing the blueprint

wormeyman:
A blueprint book whose `active_index` names a slot the book does not densely fill takes down the **whole load**, not one entry:

```
TypeError: Cannot destructure property 'upgrade_planner' of 'bps[active_index]' as it is undefined.
    at getFlattenedActiveIndex (Book.ts:94)
    at new Book (Book.ts:17)
```

`getFlattenedActiveIndex` read `bps[active_index]` with no bounds guard. Same class as #286 and #288 - an unguarded read costing the entire blueprint.

Found by PR #332's first CI run: six tests failing across three shards, one cause. Part of #329.

## `active_index` is an inventory slot, not an array position

This is the part that decided the fix, so here is the evidence trail in the order `tools/oracle/README.md` prescribes.

**Reference sources first, and they do not settle it.** No Lua under `data/base` or `data/core` reads `active_index`, and the shipped data has 89 blueprint strings but **zero** books, so there is no worked example. `doc-html/runtime-api.json` in both 2.0.77 and 2.1.14 says only *"The active blueprint index for this blueprint book"*, and nothing in either install documents the blueprint-string format at all. The changelog is suggestive but not decisive: *"Changed LuaItemStack::active_index to return nil if the blueprint book inventory has zero slots"*. Our own `blueprintSchema.json` and `types.ts` type it `integer` and say nothing.

**So it went to the binary**, which the oracle README lists as step 3 and records as having been needed by nothing until now. Both installs ship a full PDB, so the symbols are named:

- `loadBlueprintBookFromPropertyTree` (2.1.14 `0x14108d380`, 2.0.77 `0x140f5d870`) validates with `cmp dx, WORD PTR [r14+0x98]` - the book **inventory's slot count**, not any array length - then `mov WORD PTR [r14+0x118], dx`, stored verbatim with no remapping.
- `saveBlueprintBookToPropertyTree` (`0x1410897d0`) walks inventory slots, skips empty ones - which is exactly where the gaps in `index` come from - writes each entry's `index` as that slot counter, and writes `active_index` from the same `+0x118` field untransformed.
- `LuaItemCommon::luaReadActiveIndex` reads `+0x118` and `inc edx`, the 1-based Lua view the changelog mentions.

Two consequences that changed the fix:

1. An `active_index` naming an **empty** slot is legal and round-trips. The game checks capacity, not occupancy.
2. One past capacity is silently dropped to 0.

No probe was needed, and none was run.

## The reproducing blueprint

From upstream #277, added to the corpus by #332:

```
blueprint_book.active_index: 5    blueprints.length: 3
  [0] index=0  blueprint
  [1] index=1  blueprint
  [2] index=5  upgrade_planner
```

The user left gaps in the book's grid, so the planner sits in slot 5. `bps[5]` is `undefined`.

## What changed

`packages/editor/src/core/Book.ts`, three changes. The comment block carries the citation, addresses and disassembly.

- **Lookup is by slot.** `resolveActiveEntry` does `bps.findIndex(entry => entry.index === active_index)`; no match falls back to the first blueprint, which is what the game does. An earlier draft fell back to the *array position* instead - the binary evidence killed it, because that opens `EARN/power-blocks-v22-0-8.rev-1` four pages from where it belongs.
- **The correction loop** became `countNestedBlueprints(bps.slice(0, pos))`. Provably the same sum for dense books, and it also stops counting an entry that holds nothing - the schema permits a bare `{ index: n }`, and each one was shifting every later entry by one.
- **`saveBlueprint` wrote an array position** into a *nested* book's `active_index`, while `Book.serialize` renumbers only the top level. Under slot semantics that names a different entry on reload. It writes the slot now. This one is required by the change, not optional.

Also adds one bullet to `CLAUDE.md`'s invariants recording the slot semantics, so the next person does not re-derive it from a disassembler.

## One real behaviour change, and it is a correction

Measured old-vs-new across all 37 books in `test-blueprints/` plus the #277 capture: **34 resolve identically**, 2 go from throwing to 0, and one changes:

`EARN/power-blocks-v22-0-8.rev-1.txt` (`active_index: 5` against slots `[0,1,6,7,8,9,10,12,13]`) now opens on flattened page 0 rather than page 4. **Page 0 is the correct answer** per the binary - slot 5 is empty in that book.

`tests/blueprint-loading.spec.ts` renders whatever page a corpus book opens on, so it is the one spec that will now see a different blueprint. Every page of that book is already rendered by `sprite-data.spec.ts`, so a new toast is unlikely - but Playwright cannot run on the machine this was built on, so **that is the thing to watch in this PR's CI run.** `position-grid.spec.ts` calls `selectBlueprint(0)` explicitly and `book-serialize.spec.ts` / `settings-pane-book-index.spec.ts` use dense synthetic books, so none of those are affected.

## Worth knowing

`new Book({ blueprints: [] })` threw the same way, and **the committed corpus already contains a book with `active_index: 8` over six entries.** It never crashed only because `getFlattenedActiveIndex` descends into the *active* nested book alone, and nothing selects that one. The bug has been latent in the corpus for as long as the corpus has existed.

## Tests

`packages/editor/src/core/Book.test.ts` - 12 vitest cases: the exact reproducing shape, an active slot inside a gap, an index past every slot, an active planner, an active nested book with its own `active_index`, an empty book and `blueprints: undefined`, a non-numeric `active_index`, the nested-selection serialize round trip, an entry holding nothing, and four preservation cases for flattening that was already correct.

`tests/book-corpus.test.ts` - constructs every book in `test-blueprints/`, nested ones included, **in vitest with no browser**, plus two controls asserting the corpus still holds the crashing shape and the empty-active-slot shape. `tests/wire-switch-completeness.test.ts` was the precedent for importing editor source into the unit project.

Mutation-checked against `HEAD`'s `Book.ts`: 7 of the 12 unit cases fail, and the corpus test reproduces the exact CI message. Reverting only the nested-slot write fails exactly the round-trip case.

```
$ vp check
pass: All 238 files are correctly formatted
pass: Found no warnings, lint errors, or type errors in 195 files

$ vp test
 Test Files  23 passed (23)
      Tests  280 passed (280)
```

**No committed fixture moves.** No `active_index` appears in any of them, and the specs that generate them walk every page with `selectBlueprint(i)` rather than reading the book's own active index. Nothing was regenerated.

#332 rebases onto this.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -161,6 +161,12 @@ Key files:
   blueprints without a version. Do not treat a missing version as version zero.
 - Keep migrations conditional on the blueprint's declared version. Current
   Factorio can reuse an old prototype name.
+- A blueprint book's `active_index`, and the `index` on each of its entries,
+  are inventory **slots**, not positions in the `blueprints` array. A book with
+  empty slots exports a dense array with sparse `index` values, so
+  `blueprints[active_index]` is the wrong lookup and can be out of range
+  entirely. An active slot that holds nothing is legal and means the first
+  blueprint. Measured in the shipped binary; `Book.ts` carries the citation.
 - Entity accessors preserve the distinction between absent values and empty or
   zero values. Tests in `entity-accessors.spec.ts` pin this behavior.
 - Logistic filter writes must retain unknown sections and per-filter quality,
--- a/packages/editor/src/core/Book.test.ts
+++ b/packages/editor/src/core/Book.test.ts
@@ -1,0 +1,242 @@
+import { describe, expect, it } from 'vite-plus/test'
+import { Book } from './Book'
+import { IBlueprint, IBlueprintBook, IBlueprintBookEntry } from '../types'
+
+/*
+    What `active_index` resolves to, and that resolving it never throws.
+
+    `getFlattenedActiveIndex` used to index `bps[active_index]` with no bounds
+    guard, on the assumption that `active_index` is a position in the
+    `blueprints` array. It is a slot in the book's inventory grid, and a user
+    who leaves gaps in that grid exports a dense array with sparse `index`
+    values - so `bps[active_index]` was undefined and destructuring it threw.
+    `new Book` sits on `bpString.decode`'s only path with no `try` between them,
+    so that cost the whole blueprint rather than the page hint.
+
+    These are unit tests rather than Playwright specs on purpose. Everything
+    here is arithmetic over the decoded JSON: no Factorio data, no renderer and
+    no browser is involved in choosing which blueprint a book opens on.
+    `tests/book-serialize.spec.ts` covers the same conversion from the other
+    end, through the editor.
+
+    The book shapes are the measured ones. See the citation block in Book.ts for
+    where each comes from and what it settles.
+*/
+
+/** How Factorio packs a version into the `version` field. */
+const version = (main: number, major: number, minor: number, dev = 0): number =>
+    main * 2 ** 48 + major * 2 ** 32 + minor * 2 ** 16 + dev
+
+const V_1_1 = version(1, 1, 69, 2)
+const V_2_0 = version(2, 0, 55)
+
+/*
+    A blueprint with no entities, identifiable by its label. Entity-less because
+    these tests never load Factorio data: `Blueprint`'s constructor only reaches
+    FD for entities and tiles, so a labelled empty one is safe in a node
+    environment while a populated one is not.
+*/
+const blueprint = (label: string): IBlueprint =>
+    ({ item: 'blueprint', version: V_2_0, label }) as unknown as IBlueprint
+
+const book = (
+    active_index: number,
+    blueprints: IBlueprintBookEntry[] | undefined,
+    version = V_2_0
+): IBlueprintBook => ({ item: 'blueprint-book', version, active_index, blueprints })
+
+/** The label of the blueprint the book opens on. */
+const openedLabel = (data: IBlueprintBook): string | undefined =>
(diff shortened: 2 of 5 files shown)


wormeyman:
The predicted failure happened, and it was the expected shape.

`overlay-container.spec.ts` failed on shard 2, alone. The diff was **three added keys and zero removals**:

```
- Expected  -  0
+ Received  + 10
+   "boiler": Array [ 2 ],
+   "fast-splitter": Array [ -1, 1 ],
+   "steam-engine": Array [ 2 ],
```

That is a power block. `EARN/power-blocks-v22-0-8.rev-1` declares `active_index: 5` against slots `[0,1,6,7,8,9,10,12,13]`, and slot 5 is empty - verified by decoding the committed file directly - so under slot semantics it now opens on page 0 instead of array position 5, and page 0 draws boilers, steam engines and fast splitters.

The pure-addition shape is the evidence that this is a page change rather than a branch that stopped firing. A branch going quiet would show as a removal or a shortened array, never as three new keys.

Values were taken from the CI diff, not guessed. `vp check` is clean and all 280 unit tests pass.

Two things this incidentally settles:

- The corpus has been opening that book on the wrong page in every run to date, and nothing could have caught it - both readings of `active_index` produce a valid blueprint there, so no assertion would have fired either way. Only the disassembly distinguishes them.
- Mining fixture values out of a failing CI run works in practice, not just in theory. That is the same route #332 needs for its five pinned values.

wormeyman:
@claude review

PR #335, opened 2026-09-04
Make the browser suite runnable under WSL2

wormeyman:
The browser suite has never been runnable on a WSL2 machine. Launching Chromium destroys the **entire virtual machine** - `Wsl/Service/E_UNEXPECTED`, needing `wsl.exe --shutdown` to recover - not just the browser process. Four attempts out of four, at every size from the full 243-spec suite down to a single two-test spec, and the 115 MB browser download killed it several times on its own.

This makes it work.

## It is not a missing-library fault, which is what it looks like

`playwright install-deps` needs root, and on that machine `sudo -n` fails, so missing libraries are the obvious suspect. They are not the cause. Unpack the libraries by hand and the binary starts and prints its version quite happily:

```
version ok: Google Chrome for Testing 151.0.7922.34
```

The VM dies later, when a page actually renders.

## The cause is the GPU paravirtualisation path

Measured on the machine: `/dev/dxg` is present, WSLg is running (`DISPLAY=:0`, `WAYLAND_DISPLAY=wayland-0`), and there is no `.wslconfig`, so every default is on including GPU support. Chromium therefore takes the hardware GPU route, and that route crosses into the Windows GPU driver - the one layer where a guest process can take the host service down. `Wsl/Service/E_UNEXPECTED` being a **host-side** error is the tell.

Ruled out first, both measured rather than assumed: `/dev/shm` is 7.8 GB, not the 64 MB container case, and memory is 15 GiB with 13 GiB free. Neither could kill the VM anyway - they kill the process.

## The fix

`playwright.wsl.config.ts` spreads the committed config and adds software-rendering flags. **CI is untouched** - it applies only when passed explicitly, and Linux runners want the hardware path because it is faster.

```sh
export DISPLAY= WAYLAND_DISPLAY=
npx playwright test --config playwright.wsl.config.ts
```

Blanking those two variables is required rather than cosmetic: they are what attaches Chromium to WSLg in the first place.

The library workaround is documented in `CLAUDE.md` rather than scripted, since `apt-get download` needs no root:

```sh
apt-get download libnspr4 libnss3 libasound2t64
for f in *.deb; do dpkg -x "$f" ~/pw-libs; done
export LD_LIBRARY_PATH=~/pw-libs/usr/lib/x86_64-linux-gnu
```

Only those three were missing.

## Evidence, staged so a crash would say where it happened

| Stage | Result |
|---|---|
| binary starts | Chrome for Testing 151.0.7922.34 |
| `about:blank` + screenshot | **5,789-byte PNG, clean close** - the stage that used to kill the VM |
| the WebGL editor on localhost:8080 | 43,870-byte screenshot |
| `zoom-ladder.spec.ts` | 8 passed, 12.1s |
| `blueprint-round-trip.spec.ts` | 1 passed, 6.0s |

## Why this is worth committing rather than keeping local

**`blueprint-round-trip.spec.ts` passes against the committed `EXPECTED`.** It could not if software rendering changed the numbers - and it does not, because every pinned value in this repo is a model-level checksum or tally rather than a pixel comparison.

So fixtures recorded on WSL match what CI records. That makes a class of work possible that previously needed a different machine: recording the five pinned fixtures a corpus addition moves (#332 needs exactly this), and looking at a rendering change before merging it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -250,10 +250,31 @@ export LD_LIBRARY_PATH=~/pw-libs/usr/lib/x86_64-linux-gnu
 Re-run `ldd` on the Chromium binary after a Playwright upgrade; the missing set
 can grow.
 
-Measured 2026-09-04: with this, `blueprint-round-trip.spec.ts` passes against
-the committed `EXPECTED`. Software rendering does not change any pinned value,
-because they are model-level checksums and tallies rather than pixel
-comparisons - so fixtures recorded this way match what CI records.
+**Software rendering does not reproduce every spec, so know which half you are
+in before recording a fixture from a local run.** Measured 2026-09-04 against a
+clean base branch that is green in CI:
+
+| Spec                               | Local under swiftshader |
+| ---------------------------------- | ----------------------- |
+| `blueprint-round-trip.spec.ts`     | passes                  |
+| `entity-accessors.spec.ts`         | passes                  |
+| `sprite-data.spec.ts` (both cases) | passes                  |
+| `overlay-container.spec.ts`        | **fails**               |
+| `sprite-generation.spec.ts`        | **fails**               |
+
+The two failures are the same page error, and it is the renderer rather than
+the model:
+
+```
+TypeError: Failed to execute 'drawImage' on 'CanvasRenderingContext2D':
+The provided value is not of type '(CSSImageValue or HTMLCanvasElement or ...)'
+```
+
+Both specs assert `pageErrors` is empty, so they fail on that line rather than
+on a value. The specs whose pins are model-level checksums and tallies are
+unaffected and can be recorded here; **`overlay-container` and
+`sprite-generation` must be recorded from CI.** Re-check this table rather than
+assuming it, because which specs touch a canvas can change.
 
 Two rules the specs cannot enforce:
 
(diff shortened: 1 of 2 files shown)


wormeyman:
**Correction to this PR's original claim.** The body said software rendering changes no pinned value. Measured against a clean base branch that is green in CI, that is only true of half the specs.

| Spec | Local under swiftshader |
|---|---|
| `blueprint-round-trip.spec.ts` | passes |
| `entity-accessors.spec.ts` | passes |
| `sprite-data.spec.ts` (both cases) | passes |
| `overlay-container.spec.ts` | **fails** |
| `sprite-generation.spec.ts` | **fails** |

Both failures are the same page error, and it is the renderer rather than the model:

```
TypeError: Failed to execute 'drawImage' on 'CanvasRenderingContext2D':
The provided value is not of type '(CSSImageValue or HTMLCanvasElement or ...)'
```

Both specs assert `pageErrors` is empty, so they fail on that line rather than on a mismatched value.

How this was caught: the same two specs failed while recording #332's fixtures, and rather than assume the new blueprints caused it, I re-ran them on the clean base branch. They fail there too, so it is the renderer.

The practical consequence, now written into `CLAUDE.md` in this PR: the model-level specs can be recorded on WSL, and **`overlay-container` and `sprite-generation` must be recorded from CI**. That is a real limit on what this change buys, and it belongs in the file rather than being discovered by whoever records a fixture next.

The rest of the PR stands - the suite runs, the VM survives, and three of the five pinned fixtures are now recordable locally where none were before.

wormeyman:
@claude review

PR #356, opened 2026-09-04
Draw the electromagnetic plant's core instead of only its shell

HansLuft778:
draw_assembling_machine returns idle_animation.layers for a machine with always_draw_idle_animation, and the electromagnetic plant's idle animation is only its outer shell - the rotating core is a separate working_visualisation the draw never reached, so the plant renders with a hole where the core should be.

Add back the plant's one unconditional working_visualisation (always_draw, no name, no enabled_by_name). Scoped to electromagnetic-plant by name on purpose: other crafting machines' always_draw entries are apply_runtime_tint masks that read wrong without the base they tint (cryogenic-plant) or are recipe-gated (foundry), so a general pass is a wider change than this hole.

sprite-data.json's electromagnetic-plant digests move accordingly - every one gains the two core layers (one is a shadow getParts drops).

### DIFF
--- a/packages/editor/src/core/spriteDataBuilder.ts
+++ b/packages/editor/src/core/spriteDataBuilder.ts
@@ -1105,14 +1105,37 @@ function draw_artillery_wagon(
         return layers
     }
 }
+/**
+ * The electromagnetic plant's idle animation is only its shell - its core is a
+ * separate `working_visualisation` the draw never picked up, leaving a hole.
+ * Its one unconditional entry (`always_draw`, no `name`, no `enabled_by_name`)
+ * is that core.
+ *
+ * Only the electromagnetic plant. Other crafting machines' `always_draw`
+ * visualisations are `apply_runtime_tint` masks that read wrong without the
+ * base they tint (cryogenic-plant) or recipe-gated (`enabled_by_name`, foundry);
+ * folding those in is a wider change than fixing the hole this leaves.
+ */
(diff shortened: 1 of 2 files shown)


wormeyman:
THANK YOU SO MUCH this will fix cargo bay's as well!

wormeyman:
Verified this locally on x86-64 Linux before merging, and it holds up.

The predicate picks out exactly one entry. The electromagnetic plant has 10
`working_visualisations`, and only index 0 matches
`always_draw && !name && !enabled_by_name && animation`. Of the rest, four carry
names (`warm-up`, `rotation`, `rotation-continue`, `cool-down`), four lack
`always_draw`, and one has no animation, so there is nothing else the lookup
could reach by accident.

The two layers it adds are `electromagnetic-plant-main-warm-up.png` and
`electromagnetic-plant-shadow-warm-up.png`, which moves the digest from 2 to 4
and matches your fixture change exactly.

Scoping holds too. In the synthetic sweep only `electromagnetic-plant` moved;
`cryogenic-plant` (`7:eed8a241`) and `foundry` are untouched, which is what your
note about the runtime-tint masks predicted. All 5 `sprite-data.spec.ts` tests
pass with this merged onto current main.

Here it is drawn. I composited the layers from Factorio's own source PNGs rather
than screenshotting, because the machine I checked on cannot render the editor
canvas:

electromagnetic plant before and after

Thanks for explaining why the other crafting machines were left out. That saved
most of the review.

PR #369, opened 2026-09-05
Add persistent multi-entity selection, drag-to-move, and in-place mirror

HansLuft778:
Ctrl-drag already sweeps and copies, Ctrl+Right-drag already sweeps and deletes, but both throw the swept list away the moment the mouse button is released. There's no way to select a group and then decide what to do with it, and no way to move an existing group at all short of copying it, placing the copy, and deleting the originals by hand.

Alt+Left-drag sweeps a rectangle the same way copy/delete do, except the entities it covers stay selected after release. A plain Left-drag starting on one of them now picks up and moves the whole group; empty space or a non-selected entity still pans/opens its editor exactly as before. The drag itself writes nothing to the model until it's dropped - it only displaces the sprites - so a completed move is one undo step and cancelling (Escape or right-click) is a real no-op, not a remove that needs undoing back.

Shift+F/Shift+G mirror the selection in place, reusing the existing per-entity flip math rather than duplicating it. Q and Escape both clear the selection, and deleting a selected entity through the existing Ctrl+Right-drag drops it from the selection automatically.

What changed:
- New EditorMode.SELECT (Alt+Left-drag) and EditorMode.MOVE (drag a selected entity), both in BlueprintContainer
- Selection highlight uses Factorio's own multiplayer_selection cursor box, tracked independently of hover/copy/delete so it doesn't fight with them
- New PositionGrid.canGroupRelocate - canMoveTo lifts only one entity out of the grid before checking, so it wrongly refuses a group whose members would block each other (two entities shifting along their own row, for instance); the group form lifts all of them first
- New Entity.relocate, the unchecked write behind a group commit; the position setter now routes through it after its own checks
- Fixed a side effect this surfaced: AltLeft also toggles the entity-info overlay (showInfo), so an Alt-drag was flipping it on every select. The toggle now fires on Alt's key-up and is skipped when a drag happened during the hold - tapping Alt still shows it, holding it to drag doesn't
- vp check clean, vp test 300/300 (new groupRelocation.test.ts included), 12 new Playwright cases in persistent-selection.spec.ts plus the existing mode/keybind specs re-run clean

### DIFF
--- a/packages/editor/src/Editor.ts
+++ b/packages/editor/src/Editor.ts
@@ -188,12 +188,49 @@ export class Editor {
 
     /**
      * The interaction mode the canvas is in, by name - NONE, EDIT, PAINT, PAN,
-     * COPY or DELETE. Exposed for tests/editor-mode-input.spec.ts, which is the
-     * first spec to drive real pointer and keyboard input (issue #44); the name
-     * rather than the enum member so a test does not have to import it.
+     * COPY, DELETE, SELECT or MOVE. Exposed for tests/editor-mode-input.spec.ts,
+     * which is the first spec to drive real pointer and keyboard input (issue
+     * #44); the name rather than the enum member so a test does not have to
+     * import it.
      */
     public get mode(): string {
         return EditorMode[G.BPC.mode]
+    }
+
+    /**
+     * The entity numbers in the persistent selection, in insertion order. The
+     * only way a spec can see what an Alt-drag swept, since the selection is
+     * not a mode - `mode` reads NONE once the drag is released. See
+     * tests/persistent-selection.spec.ts.
+     */
+    public get selectedEntityNumbers(): number[] {
+        return G.BPC.selectedEntityNumbers
+    }
+
+    /** Whether the entity's selection box is tinted as blocked. See tests/persistent-selection.spec.ts. */
+    public selectionHighlightBlocked(entityNumber: number): boolean {
+        return G.BPC.selectionHighlightBlocked(entityNumber)
+    }
+
+    /**
+     * Whether the entity-info overlay (the AltLeft toggle) is showing. Needed
+     * to prove a tap of Alt still toggles it while an Alt-drag selection does
+     * not - see `BlueprintContainer.consumeAltUsedForDrag`. See
+     * tests/persistent-selection.spec.ts.
+     */
+    public get infoOverlayVisible(): boolean {
+        return G.BPC.infoOverlayVisible
+    }
+
+    /**
+     * `History.revision`. A group move or mirror has to be *one* undo step, and
+     * counting transactions is the only way a spec can tell one from two - the
+     * end state of the entities is the same either way. See
+     * tests/persistent-selection.spec.ts, and `revision`'s own doc comment for
+     * what this number does and does not promise.
+     */
+    public get historyRevision(): number {
+        return G.bp.history.revision
     }
 
     /**
@@ -515,6 +552,19 @@ export class Editor {
                     onRelease: () => G.BPC.exitDeleteMode(),
                 },
             },
+            // any -> SELECT; the swept entities stay selected after release
+            selectGroup: {
+                trigger: {
+                    button: MouseButton.Left,
+                },
+                modifiers: {
+                    alt: true,
+                },
+                callbacks: {
+                    onPress: () => G.BPC.enterSelectMode(),
+                    onRelease: () => G.BPC.exitSelectMode(),
+                },
+            },
 
             moveUp: {
                 trigger: {
@@ -552,14 +602,22 @@ export class Editor {
                     onRelease: () => G.BPC.moveEnd('right'),
                 },
             },
+            // A tap of Alt shows the overlay; holding it to Alt-drag a
+            // selection (selectGroup, below) must not also toggle it. Both
+            // bind plain AltLeft with no way to tell them apart at press time
+            // - `enterSelectMode` cannot know yet whether this Alt is headed
+            // for a drag - so the toggle itself waits for the key-up, and is
+            // skipped exactly when a selection sweep happened in between.
             showInfo: {
                 trigger: {
                     code: 'AltLeft',
                 },
                 callbacks: {
-                    onPress: () => {
-                        G.BPC.overlayContainer.toggleEntityInfoVisibility()
-                        return true
+                    onPress: () => true,
+                    onRelease: () => {
+                        if (!G.BPC.consumeAltUsedForDrag()) {
+                            G.BPC.overlayContainer.toggleEntityInfoVisibility()
+                        }
                     },
                 },
             },
@@ -569,7 +627,12 @@ export class Editor {
                 },
                 callbacks: {
                     onPress: () => {
-                        Dialog.closeLast()
+                        // A dialog is on top of the canvas, so it goes first.
+                        if (Dialog.anyOpen()) {
+                            Dialog.closeLast()
+                        } else {
+                            G.BPC.escape()
+                        }
                         return true
                     },
                 },
--- a/packages/editor/src/containers/BlueprintContainer.ts
+++ b/packages/editor/src/containers/BlueprintContainer.ts
@@ -15,6 +15,8 @@ import { Tile } from '../core/Tile'
 import { Entity } from '../core/Entity'
 import { Blueprint } from '../core/Blueprint'
 import { IConnection } from '../core/WireConnections'
+import { GroupRelocation } from '../core/PositionGrid'
+import U from '../core/generators/util'
 import { stepZoom, WheelZoom } from '../core/zoomLevels'
 import { IPoint } from '../types'
 import { Dialog } from '../UI/controls/Dialog'
@@ -70,7 +72,14 @@ export enum EditorMode {
     COPY,
     /** Active when selecting multiple entities for deletion */
     DELETE,
+    /** Active while Alt+dragging the rectangle that becomes the persistent selection */
+    SELECT,
+    /** Active while dragging the persistent selection to a new place */
+    MOVE,
 }
+
+/** The live rectangle's colour in SELECT mode - not copy's green nor delete's red. */
+const SELECT_AREA_COLOR = 0x3399ff
 
 export class BlueprintContainer extends Container {
     private static _moveSpeed = 10
@@ -174,6 +183,42 @@ export class BlueprintContainer extends Container {
     private copyModeUpdateFn: ((endX: number, endY: number) => void) | undefined
     private deleteModeUpdateFn: ((endX: number, endY: number) => void) | undefined
     private copySettingsActive = false
+
+    /*
+        The persistent selection: what an Alt-drag swept, kept past the mouse-up
+        so it can be dragged somewhere else or mirrored. The only entity list
+        here that outlives a gesture - copyModeEntities and deleteModeEntities
+        above are emptied on the way out of their modes.
+    */
+    private readonly selectedEntities = new Set<Entity>()
+    /*
+        Set the moment an Alt-drag actually starts a selection sweep, and read
+        (and cleared) once, on the matching AltLeft key-up - see
+        `consumeAltUsedForDrag`. What lets `showInfo` (also bound to plain
+        AltLeft) tell a tap of Alt from an Alt held to select: it defers its
+        toggle to the key-up and skips it exactly when this is true.
+    */
+    private altUsedForDrag = false
+    /*
+        Per member, the function that detaches the listeners keeping it
+        selected: the ones that move its box when it moves and drop it from the
+        set when it is destroyed by anything else - a delete drag, a mine. Every
+        way out of the selection goes through deselect(), which runs and forgets
+        this.
+    */
+    private readonly selectionListeners = new Map<Entity, () => void>()
+    /* The SELECT-mode pair, same shape and lifetime as the copy/delete ones above. */
+    private selectModeEntities: Entity[] = []
+    private selectModeUpdateFn: ((endX: number, endY: number) => void) | undefined
+    /*
+        Present only during a MOVE drag. `moved` flips on the first tile
+        crossed, and is what separates a drag from a click on a selected
+        entity: a click still opens the entity's editor, as it does anywhere.
+    */
+    private moveDrag:
+        | { entities: Entity[]; pressed: Entity; delta: IPoint; moved: boolean }
+        | undefined
+    private moveDragUpdateFn: ((endX: number, endY: number) => void) | undefined
 
     // PIXI properties
     public readonly eventMode = 'static'
@@ -287,6 +332,15 @@ export class BlueprintContainer extends Container {
                 this.viewport.translateBy(e.movement.x, e.movement.y)
             },
             panStart: (): boolean => {
+                /*
+                    First, because it cannot be a sibling action: `pan` is the
+                    first plain-Left, zero-modifier action registered, and
+                    ActionRegistry tries equal-modifier actions in registration
+                    order and stops at the first that answers true - which this
+                    does for every press in NONE. A "drag the selection" action
+                    registered anywhere else would never be reached.
+                */
+                if (this.enterMoveMode()) return true
                 if (this.mode === EditorMode.NONE) {
                     this.cursor = 'move'
                     this.setMode(EditorMode.PAN)
@@ -296,6 +350,10 @@ export class BlueprintContainer extends Container {
                 return false
             },
             panEnd: (): void => {
+                if (this.mode === EditorMode.MOVE) {
+                    this.exitMoveMode()
+                    return
+                }
                 if (this.mode === EditorMode.PAN) {
                     this.off('globalpointermove', panModule._onPan)
                     this.setMode(EditorMode.NONE)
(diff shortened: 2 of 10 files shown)


wormeyman (commented):
A review of the whole branch at `ce3dbae`. Ten findings are inline below; each one was traced through the code rather than pattern matched, and the line numbers are this branch's.

Seven of them share a cause. A drag holds a snapshot of the selection, and both undo and a modified press can run while that snapshot is live. Neither one tells the drag, so the editor can end up stuck in MOVE mode for the rest of the session, or holding a selected entity that no longer exists and can never be cleared.

The three I would fix first:

1. Undo during a MOVE drag leaves the editor stuck in MOVE, throwing on every later click and Escape (`BlueprintContainer.ts:1145`).
2. A modified press during a drag overwrites the mode without exiting it, so `moveDrag`, its grid subscription and every sprite offset leak (`:1000`, and the same guard shape in `enterDeleteMode` and `enterCopyMode`).
3. `mirrorSelection` commits the move before writing the new direction, so the position grid and the model disagree for any entity whose footprint transposes under a flip (`:1281`).

Five cleanup items were cut from the inline list to keep it readable:

- `PositionGrid.canGroupMoveTo` (`:452`) has no callers anywhere in the tree.
- `canGroupRelocate` removes and re-adds every member on the grid for every pointer tile of a drag. An `ignore` set on `isAreaAvailable` would avoid the writes.
- Each `relocate` in the commit transaction triggers a full sprite re-sort, so dropping n entities costs n sorts of the whole blueprint. `initBP` (`:1470`) shows the batching pattern.
- The marquee in `enterSelectMode` is a third verbatim copy of the copy and delete ones.
- The wire-reach rule now has a sixth copy. One predicate on `WireConnections` would serve the setter, `connectionsReach` and `canRelocateGroup`.

packages/editor/src/containers/BlueprintContainer.ts
```diff
@@ -814,6 +898,397 @@ export class BlueprintContainer extends Container {
...
+        const drag = this.moveDrag
+        this.moveDrag = undefined
+        if (this.moveDragUpdateFn !== undefined) {
+            this.gridData.off('update32', this.moveDragUpdateFn, this)
+            this.moveDragUpdateFn = undefined
+        }
+        if (drag === undefined) throw new Error('MOVE mode with no drag in progress')
+
+        for (const e of drag.entities) {
+            const container = EntityContainer.containerOf(e.entityNumber)
```
wormeyman, line 1145: **Undo during a MOVE drag leaves the editor stuck in MOVE mode.**

`moveDrag.entities` is a snapshot taken at `:1093`, and undo is bound with no mode or button gate (`Editor.ts:755`). When an undo destroys a member, `EntityContainer.mappings` drops the entry and the `destroy` listener prunes `selectedEntities`, but nothing prunes that snapshot.

Repro: paint a chest, Alt-sweep it with two others, left-press one, drag a tile, and press Ctrl+Z while still holding. The next tile runs `previewMove`, where `containerOf` throws (`EntityContainer.ts:29-34`). On release, `exitMoveMode` has already cleared `moveDrag` at `:1137` and then throws here, so it never reaches `setMode(EditorMode.NONE)` at `:1151`. From then on every click, Escape and mine re-enters `exitMoveMode` and throws "MOVE mode with no drag in progress". The other dragged sprites keep a non-zero drag offset the whole time.

Fix: set NONE before the restore loop and skip entities `EntityContainer.mappings` no longer has, or prune `moveDrag.entities` in `deselect`.

packages/editor/src/containers/BlueprintContainer.ts
```diff
@@ -814,6 +898,397 @@ export class BlueprintContainer extends Container {
...
+
+    /**
+     * Alt+Left-drag: the copy-mode marquee, except that on release the swept
+     * entities *stay* selected instead of being handed to a paint container.
+     * The box each one gets while the rectangle covers it is the persistent
+     * one, so nothing changes visually on release - it simply stops being
+     * cleared. A new sweep replaces the previous selection outright.
+     */
+    public enterSelectMode(): boolean {
+        if (this.mode === EditorMode.SELECT) return false
```
wormeyman, line 1000: **This guard shape lets a modified press overwrite MOVE without exiting it.**

`enterSelectMode` here, `enterDeleteMode` (`:842`) and `enterCopyMode` (`:783`) each refuse only their own mode and PAINT. None of them refuse MOVE, so a modified press during a live drag calls `setMode` straight over it.

Repro: Alt-select A and B, left-press A and drag a tile, then press Ctrl+Right while still holding left. `sortedActions` orders by modifier count (`actions.ts:209-210`), so `deleteSelection` (`Editor.ts:543`, one modifier) runs ahead of the plain `mine`, and the new MOVE guard in `mineStart` at `:466` never gets a chance. The red sweep sets DELETE and its release sets NONE. The left release then reaches `panEnd` at `:353`, sees NONE, and skips `exitMoveMode`, which would early-return on `mode !== MOVE` at `:1135` anyway.

What leaks: `moveDrag` stays set, `moveDragUpdateFn` stays subscribed to `gridData` "update32" (`:1100`) and the next `enterMoveMode` overwrites the reference so it can never be unsubscribed, and every sprite keeps its drag offset. Pointer moves keep sliding A away from its model position for the rest of the session, and if the sweep deleted B, `previewMove` throws. `selectModeUpdateFn` leaks the same way from this method.

Fix: have each `enter*` exit or refuse MOVE first.

packages/editor/src/containers/BlueprintContainer.ts
```diff
@@ -814,6 +898,397 @@ export class BlueprintContainer extends Container {
...
+            G.logger({ text: 'The mirrored selection does not fit.', type: 'warning' })
+            return
+        }
+
+        this.bp.history.transaction(
+            vertical ? 'Mirror selection vertically' : 'Mirror selection horizontally',
+            () => {
+                for (let i = 0; i < flipped.length; i++) {
+                    const { entity, copy } = flipped[i]
+                    entity.relocate(targets[i].position)
```
wormeyman, line 1281: **`relocate` runs before the new direction is written, so the grid and the model disagree.**

`canGroupRelocate` validated the post-flip footprint, but `relocate`'s `onDone` runs synchronously and sizes `removeTileData` / `setTileData` from `entity.size` (`Entity.ts:316-317`), which still reads the old direction. The direction setter writes the raw field and emits (`Entity.ts:357-363`); it never touches the grid.

That only matters when a flip transposes the footprint, and it does. `getEntitySize` classes direction with `Math.floor((dir % 8) / 4) * 4` (`factorioData.ts:627`). Select a `curved-rail-a` at direction 2 (2x6) and press Shift+G: the flip maps 2 to 6, which lands in the 6x2 class. The grid gets the 2x6 cells, then the next `removeTileData` uses 6x2 and leaves stale cells behind, which is the drift `PositionGrid.entityAt` exists to throw on. `railgun-turret` (3x5, 8-way) crosses the same way, and `getFlippedCopy` refuses only train stops and the two rail signals, so the path is open.

Fix: write the direction before `relocate` with the grid entry removed across both writes, or give `PositionGrid` a relocate that takes position and direction together.

packages/editor/src/containers/BlueprintContainer.ts
```diff
@@ -814,6 +898,397 @@ export class BlueprintContainer extends Container {
...
+        }
+
+        this.setMode(EditorMode.NONE)
+        this.updateHoverContainer()
+
+        for (const e of this.selectModeEntities) {
+            if (cancel) {
+                this.overlayContainer.hideSelectionHighlight(e.entityNumber)
+            } else {
+                this.select(e)
```
wormeyman, line 1058: **An entity destroyed since the last sweep update can never leave the selection.**

`selectModeEntities` is recomputed only on "update32". If a swept entity was destroyed since then, `select(e)` adds it to `selectedEntities` at `:960` and `redrawBox` then throws on `containerOf` at `:966`, before `selectionListeners.set` runs at `:976`. `deselect` looks up that listener entry first and returns early when it is missing (`:985`), so Escape, Q and `clearSelection` all leave the dead entity in the selection.

Repro: hold Alt+Left with the rectangle over a freshly painted chest and press Ctrl+Z. `hasModifiers` (`actions.ts:445`) ignores the extra Alt, so undo fires. Release without moving the pointer. The throw also aborts this loop, so entities later in the sweep are never selected and `selectModeEntities = []` at `:1061` never runs. `flip()` keeps routing Shift+F and Shift+G to `mirrorSelection`, which throws on `containerOf` again.

Fix: skip entities `EntityContainer.mappings` no longer has, or add to `selectedEntities` only after `redrawBox` succeeds.

packages/editor/src/containers/BlueprintContainer.ts
```diff
@@ -688,6 +769,9 @@ export class BlueprintContainer extends Container {
         }
         this.exitCopyMode(true)
         this.exitDeleteMode(true)
+        this.exitSelectMode(true)
+        // Q is a second way out of a selection, alongside Escape.
+        this.clearSelection()
```
wormeyman, line 774: **Q during a drag silently commits the move, and Q on a hovered entity throws the selection away.**

In MOVE mode the EDIT and PAINT branches above are skipped and `exitSelectMode` early-returns, so this `clearSelection()` hides every box and detaches every listener while `moveDrag` stays live. The rest of the drag gives no feedback, and the release still runs `exitMoveMode`, which commits "Move selection". Escape during a drag cancels the move; Q commits it.

The second case: with a selection active, hover an unselected entity and press Q. The EDIT branch spawns the paint container and then falls through to this line, while the quickbar and Ctrl+V callers of `spawnPaintContainer` (`Editor.ts:436`, `:653`) leave the selection alone. The only Q test moves the pointer to empty space, so neither path is covered.

Fix: call `exitMoveMode(true)` first, and clear only in the branch that means to.

packages/editor/src/containers/BlueprintContainer.ts
```diff
@@ -814,6 +898,397 @@ export class BlueprintContainer extends Container {
...
+
+        this.bp.history.transaction(
+            vertical ? 'Mirror selection vertically' : 'Mirror selection horizontally',
+            () => {
+                for (let i = 0; i < flipped.length; i++) {
+                    const { entity, copy } = flipped[i]
+                    entity.relocate(targets[i].position)
+                    if ((copy.rawEntity.direction ?? 0) !== (entity.rawEntity.direction ?? 0)) {
+                        entity.direction = copy.rawEntity.direction ?? 0
+                    }
```
wormeyman, line 1284: **Mirrored splitters keep the wrong lane priority when they face south or west.**

`getFlippedCopy` swaps the input and output priority only when the new direction lands in {4, 8} for a vertical flip or {0, 12} for a horizontal one (`Entity.ts:1300-1305`). Since the new direction is `(axisDir * 2 - old) % 16`, that condition is true only when the old direction was 0 or 4. A reflection always reverses chirality, so the swap should be unconditional. Under a vertical flip, N and E swap correctly, S(8) to N(0) keeps "left" when "left" now names the opposite lane, and W(12) to W keeps it too. A horizontal flip mirrors the same two cases, and flipping twice does not return to the start.

Repro: select a south-facing splitter with output priority "left" and press Shift+G. The committed entity faces north with its priority on the west lane, the mirror image of where it belongs.

This predates the PR in the paint-flip caller, but this is the first path that writes it onto committed blueprint entities (`:1284-1285`), and no spec covers a flipped splitter.

Fix: swap unconditionally in `getFlippedCopy`.

packages/editor/src/containers/BlueprintContainer.ts
```diff
@@ -814,6 +898,397 @@ export class BlueprintContainer extends Container {
...
+     * entities *stay* selected instead of being handed to a paint container.
+     * The box each one gets while the rectangle covers it is the persistent
+     * one, so nothing changes visually on release - it simply stops being
+     * cleared. A new sweep replaces the previous selection outright.
+     */
+    public enterSelectMode(): boolean {
+        if (this.mode === EditorMode.SELECT) return false
+        if (this.mode === EditorMode.PAINT) this.painting.destroy()
+
+        this.altUsedForDrag = true
```
wormeyman, line 1003: **`altUsedForDrag` is set by any `selectGroup` press but cleared only by an AltLeft key-up.**

`ActionRegistry.setModifiers` keys on `e.key === 'Alt'` (`actions.ts:222`, `:262`), so the right Alt satisfies `selectGroup`'s modifiers and sets this flag. `showInfo` binds code "AltLeft" (`Editor.ts:612`) and `triggerMatches` compares `e.code` (`actions.ts:483`), so an AltRight release never reaches `consumeAltUsedForDrag`. The next AltLeft tap finds the stale `true`, resets it, and skips the toggle, so the tap does nothing. Nothing else clears the flag: not `exitSelectMode`, not `releaseAll`, not `escape`.

Rebinding `selectGroup` off Alt does the same thing permanently. Keybinds are editable (`packages/website/src/settingsPane.ts:189`) and persisted through `importKeybinds` (`packages/website/src/index.ts:951`), so after a rebind every sweep eats the next Alt tap.

Fix: set the flag only when the held modifier really is Alt, or clear it in `exitSelectMode` when Alt is not down.

packages/editor/src/Editor.ts
```diff
@@ -552,14 +602,22 @@ export class Editor {
...
             showInfo: {
                 trigger: {
                     code: 'AltLeft',
                 },
                 callbacks: {
-                    onPress: () => {
-                        G.BPC.overlayContainer.toggleEntityInfoVisibility()
-                        return true
+                    onPress: () => true,
+                    onRelease: () => {
```
wormeyman, line 617: **Moving the toggle to key-up loses it whenever the key-up is filtered.**

`keyup` returns early for events targeted at an input or textarea (`Editor.ts:897-901`), so `releaseKey` never runs for them. Hold Alt on the canvas, click into the Import/Export textarea or any dialog input, and release Alt there: `activeRelease` stays pending. `Action.press` overwrites it without invoking it (`actions.ts:463-469`), so the next Alt tap is eaten. If the user then Alt-Tabs away, `releaseAll` on window blur (`Editor.ts:903`) fires the stale callback and the overlay toggles on its own. On the base branch the same sequence toggled on the first press.

Fix: either let the keyup filter through for modifier keys, or invoke a pending `activeRelease` when a new press finds one.

packages/editor/src/containers/BlueprintContainer.ts
```diff
@@ -814,6 +898,397 @@ export class BlueprintContainer extends Container {
...
+     * members; false - and nothing done - otherwise, so `panStart` carries on
+     * to pan exactly as it did before selections existed. A press on the
+     * background or on an entity that is not selected is therefore still a pan.
+     *
+     * Nothing is written to the model from here until `exitMoveMode` commits:
+     * the drag previews by displacing the real sprites (`EntityContainer.
+     * setDragOffset`) and the selection boxes, and a cancel just puts them
+     * back. That is what makes cancelling a true no-op and a completed move a
+     * single undo step - the reasons a detach-and-carry design was rejected
+     * (docs/superpowers/specs/2026-09-05-persistent-selection-design.md).
```
wormeyman, line 1075: **This design spec is not in the PR or on the base branch.**

`docs/superpowers/specs/2026-09-05-persistent-selection-design.md` is cited here, in `packages/editor/src/core/groupRelocation.test.ts:9`, and in `tests/persistent-selection.spec.ts:10`. `git ls-tree -r` on this branch finds no such file, and `docs/superpowers/specs` still holds the six specs `CLAUDE.md` counts. A reader following any of the three pointers finds nothing.

Fix: commit the spec and bump the count in `CLAUDE.md` to 7, or inline the reasoning at each of the three sites.

tests/persistent-selection.spec.ts
```diff
@@ -0,0 +1,320 @@
...
+    await openEditorWithChests(page)
+    await altSelect(page, 1, 2)
+
+    const chest = await screenOf(page, 1)
+    await page.mouse.move(chest.x, chest.y)
+    await page.mouse.down()
+    expect(await modeOf(page)).toBe('MOVE')
+    await page.mouse.up()
+    expect(await dialogs(page)).toBe(1)
+    expect(await positionOf(page, 1)).toEqual(await positionOf(page, 1))
```
wormeyman, line 196: **This compares a value to itself, so the "without dragging" half of the test cannot fail.**

The test captures no position and no revision before `mouse.down()`, so its only real assertions are the mode and the dialog count. A regression that teleported the chest or wrote a history entry on a plain click would still pass.

The neighbouring tests show the pattern: "Escape mid-drag" at `:203` and the occupied-tile case at `:227` both capture `positionOf` and the revision before the press and compare after `mouse.up()`.

wormeyman:
I re-checked the three blocking items by driving the branch merged onto the current base, and opened HansLuft778/factorio-blueprint-editor#1 against `feature/persistent-selection` with fixes and tests for items 1 and 3. Merging it there updates this PR.

Item 2 turned out not to be reachable from a mouse: a second button pressed while one is held fires `pointermove`, not `pointerdown`, so no `enter*Mode` runs during a held drag. The guard is still missing but nothing gets to it. Item 3 is worse than the review said: after a Shift+G on a curved rail, deleting it and hovering one of the stale cells throws `Position grid references entity 1`. Details and numbers are in that PR.

Issue #187, opened 2026-08-05
Factorio 2.1: regenerate data.json and sprites, and move the dependencies pinned to 2.0, once it lands on stable


wormeyman:
## The shape

The exporter is pinned to **2.0.68** at `packages/exporter/src/main.rs:13`:

```rust
static FACTORIO_VERSION: &str = "2.0.68";
```

Everything downstream is a snapshot of that install. When 2.1 reaches stable, the
committed output stops describing the game the editor claims to render:

- `packages/exporter/data/output/data.json` - 155 entities, 340 items, 653
  recipes, 22 tiles, 155 signals, 33 fluids, 569 utility sprites
- **8,776 committed `.basis` files**, 458 MB of tracked output

New 2.1 prototypes are not a soft failure. `stripUnknownPrototypes` drops any
entity or tile `FD` does not have, so a 2.1 blueprint carrying a new prototype
loads with that entity silently missing and a warning - the same path issue #46
covered for tiles. Existing entities that gain or change sprite fields are worse,
because they fail inside `generateGraphics()` and degrade to a placeholder box
through `getSpriteData`'s try/catch.

## This is not the same as #155

#155 regenerates from **2.0.77** - the same major line, a mechanical refresh, and
the worked example for how to run one in an isolated profile. This is the version
boundary underneath it: the point at which held dependencies have to move together
and behaviour actually differs, not just data.

## What has to move with it

Each of these is recorded in `CLAUDE.md` as a deliberate hold, and every one of
them is scoped to 2.0:

- **`typed-factorio` comes off the `factorio-2.0` dist-tag.** Upstream publishes
  a dedicated tag: `{latest: 4.2.0, factorio-2.0: 3.36.0}`, where 3.36.0 describes
  Factorio 2.0.75 and 4.2.0 describes 2.1.9. The project declares `^3.36.0` in
  both the root and `packages/editor/package.json`. It is consumed as **ambient
  types** via the tsconfig `types` array with **zero import sites**, so a major
  changes global types everywhere at once with nothing to stage the migration
  through - and `strict: true` is on across every package, so the error count
  lands all at once too.
- **`.npmrc`'s `legacy-peer-deps=true` needs re-testing.** It is load-bearing for
  `typed-factorio`'s non-optional peers on `lua-types` and `typescript-to-lua`,
  and v4 declares the same peers, so the conflict does not go away on its own.
- **basisu stays coupled.** Encoder and bundled transcoder must match
[...]

PR #370, opened 2026-09-05
Fix diagonal flamethrower turret rendering for Factorio 2.1

### LINKED ISSUE #366: 2.1 makes flamethrower turrets 8-way, which will break draw_fluid_turret on diagonals
Not a defect today. It becomes one the moment `data.json` is regenerated from
2.1, so it should be on the migration checklist rather than discovered
afterwards.

## The change

Factorio 2.1 changelog, readable at
`/mnt/v/factorio-2.1.14/data/changelog.txt`:

> Flamethrower turrets can now be rotated 8 ways instead of 4.

## Why that breaks rendering

`flamethrower-turret` is a `fluid-turret` and goes to `draw_fluid_turret`
(`packages/editor/src/core/spriteDataBuilder.ts:1623` at `c79c279b`), which
reaches `util.getDirName`. **`getDirName` throws on the four diagonal
directions.** `getSpriteData` catches it and substitutes
`SPRITE_GENERATION_FAILED`, so a diagonally-placed flamethrower turret would
render as a placeholder rather than a turret.

In the committed 2.0 data the entity cannot face a diagonal, so nothing reaches
the throw:

```
flamethrower-turret.folded_animation keys: east, north, south, west
  (same for preparing_animation, prepared_animation, attacking_animation)
```

Once the data carries eight keys, every diagonal placement fails.

## This is the bug #357 just fixed, one entity over

#357 hit exactly this on `railgun-turret`, the only 8-way entity in the 2.0
data. Measured there: with the fix reverted, all four diagonals returned
`FAILED`. The fix was to index the underscored `RotatedAnimation8Way` keys
directly instead of going through `getDirName`, and
`tests/railgun-turret-diagonal.spec.ts` pins it.

So the fix here is known in advance, and the spec to copy already exists. What
is not yet known is which shape 2.1 gives the animation, underscored
`north_east` keys like the railgun or something else. That wants reading off
the regenerated data rather than guessing.

## What to do

Add to the 2.1 migration checklist in #187:

1. After regenerating, check whether `flamethrower-turret.folded_animation` has
   eight keys and what they are named.
2. If so, move `draw_fluid_turret` off `getDirName` the way #357 moved
   `draw_railgun_turret`.
3. Add a diagonal spec modelled on `tests/railgun-turret-diagonal.spec.ts`, and
   confirm it fails without the fix before believing it.

Worth a wider sweep at the same time: around 20 call sites still reach
`getDirName`, and none has been audited for which of its entities can face a
diagonal. 2.1 is the release that turns that audit from theoretical into
necessary. See the "Known limits" entry in `CLAUDE.md` (#363).

byalex33:
Factorio 2.1 diagonal flamethrower turrets hit cardinal-only lookups in both their base/folded animations and pipe covers, producing SPRITE_GENERATION_FAILED. Resolve the underscored eight-way animation keys and use alternate pipe positions/directions for cover placement and connection detection.

Animation keys and alternate connection coordinates were verified against wube/factorio-data 2.1.14, base/prototypes/entity/fire.lua. Generated assets remain unchanged; the regression augments the committed 2.0 prototype with the verified 2.1 structure and distinct diagonal sprite markers.

Validation:
- Regression fails against the original renderer and passes with the fix; covers all eight facings, cardinal compatibility, alternate cover positions, and connections to diagonal neighbours.
- All 224 editor tests pass on the current upstream base.
- Formatting, lint, and type checks pass for all three changed files.
- Broader checks on the original checkout found unrelated formatting failures and a rail fixture test comparing CRLF with LF. Full 2.1 asset regeneration and visual validation remain migration work.

Closes #366

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved eight-direction rendering for fluid and flamethrower turrets, including diagonal orientations.
  * Corrected alternate pipe-cover selection based on neighboring connections.
  * Preserved clear error handling for unsupported diagonal entities.
  * Improved railgun base visualisation consistency across directions.


packages/editor/src/core/spriteDataBuilder.ts
```diff
@@ -325,6 +326,32 @@ function duplicateAndSetPropertyUsing<K0 extends SpriteNumberKey, K1 extends Spr
...
 
+function fluidConnection(
+    connection: PipeConnectionDefinition & {
+        alt_position?: IPoint | readonly [number, number]
+        alt_direction?: number
+    },
+    entityDir: number
+): { offset: IPoint; dir: number } {
+    // 2.1 alternate connections are defined at north-east, then rotate in quarter turns.
+    const diagonal = entityDir % 4 === 2
```
wormeyman: `diagonal` is a bare arithmetic test on the direction. It is not a check that
the prototype declares the 2.1 alternate fields, so **every** fluid-box entity at
direction 2, 6, 10 or 14 takes this branch - including all the ones in the
committed 2.0 data that have no `alt_position` at all. Those fall through to
`connection.position` and get rotated by `entityDir - 2`, which is the previous
cardinal.

Measured on this branch against the committed `data.json`, drawing each entity at
directions 0, 2, 4 and 6:

| entity | base `03650118` | this branch |
| --- | --- | --- |
| `storage-tank` d2 | FAILED | 6 layers |
| `storage-tank` d6 | FAILED | 6 layers |
| `pump`, `boiler`, `chemical-plant`, `oil-refinery`, `flamethrower-turret` | FAILED | FAILED |

Only `storage-tank` changes, because it is the one entity here whose own `draw_*`
does not reach `getDirName` and throw first. Its diagonal output is byte-identical
to north at direction 2 and to east at direction 6, so a diagonal storage tank now
renders as an ordinary tank with its pipe covers in the wrong place, instead of
showing the placeholder.

Direction 2 on a storage tank is invalid data, and a blueprint string can carry
it. `spriteShape.ts`'s own header is explicit that malformed data should keep
throwing so `getSpriteData` can log it, because "silent empty returns would hide
that diagnostic" - and a silently wrong render hides it just as well.

Gating on the fields being present keeps the diagnostic for everything else, and
leaves the diagonal turret path exactly as this PR intends:

```ts
const diagonal =
    entityDir % 4 === 2 &&
    (connection.alt_position !== undefined || connection.alt_direction !== undefined)
```

packages/editor/src/core/spriteDataBuilder.ts
```diff
@@ -401,13 +424,11 @@ function checkFluidConnection(x: number, y: number, entity: Entity, relDir: numb
...
     for (const connection of connections) {
-        const offset = connection.position
-            ? util.rotatePointBasedOnDir(connection.position, entity.direction)
-            : util.Point(need(connection, 'positions')[entity.direction / 4])
+        const { dir, offset } = fluidConnection(connection, entity.direction)
         if (
             x === Math.floor(entity.position.x + offset.x) &&
             y === Math.floor(entity.position.y + offset.y) &&
-            (entity.direction + need(connection, 'direction')) % 16 === (relDir + 8) % 16
+            dir === (relDir + 8) % 16
```
wormeyman, line 440: This is the other half of the change, and nothing currently pins it. Replacing the
whole direction comparison with `true`:

```ts
x === Math.floor(entity.position.x + offset.x) &&
y === Math.floor(entity.position.y + offset.y) &&
true
```

leaves the entire vitest suite green - 301 tests across 26 files, `fluidTurret.test.ts`
included. That is the scope I measured; I did not run the Playwright specs.

The neighbour case at the end of the new test only ever supplies a neighbour that
matches, so "some connection matched" and "the right connection matched" look
identical to it. A negative control closes the gap: the same neighbour with its
`alt_direction` pointing the wrong way, or sitting one tile off, must still get a
cover.

packages/editor/src/core/spriteDataBuilder.ts
```diff
@@ -1621,10 +1642,23 @@ function draw_elevated_straight_rail(
     return draw_elevated_rail(e as unknown as RailPrototype)
 }
 function draw_fluid_turret(e: FluidTurretPrototype): (data: IDrawData) => readonly SpriteData[] {
-    return (data: IDrawData) => [
-        ...baseVisualisationLayers(need(e, 'graphics_set', 'base_visualisation'), data.dir),
-        ...dirLayers(e.folded_animation, util.getDirName(data.dir)),
-    ]
+    return (data: IDrawData) => {
+        // RotatedAnimation8Way uses underscores, unlike rail sprite directions.
```
wormeyman: This array is already in this file. `ROTATED_8WAY_KEYS` at line 1005 holds the same
eight strings in the same order, and `draw_railgun_turret` at line 1018 uses it for
this exact bug one entity over (#357). A third copy gives the next 8-way entity two
tables to keep in step.

```ts
const key = ROTATED_8WAY_KEYS[(data.dir >> 1) & 7]
```

That also settles the indexing just below. `data.dir / 2` gives a fractional index
for an odd direction, so the lookup returns `undefined` and `layersOf` throws with a
message naming neither the entity nor the direction; `(data.dir >> 1) & 7` is total.
Both end in the placeholder today, so this is not a behavior change - the better
form is simply already in the file.

While you are in here: `draw_railgun_turret` still unwraps `base_visualisation` by
hand in the three lines that `baseVisualisationLayers` now does for it. Worth
collapsing, so the two turret draws stay the same shape.

packages/editor/src/core/fluidTurret.test.ts
```diff
@@ -0,0 +1,121 @@
+import { expect, it } from 'vite-plus/test'
+import exportedData from '../../../exporter/data/output/data.json'
```
wormeyman: Two smaller notes, neither blocking.

This is the first test in the editor project to `import` the generated `data.json`.
`vite.config.ts` (lines 239-243) explains that project is deliberately narrowed, and
that "a test that reads a file off disk therefore has to sit under a directory the
root config owns". A JSON import is not an `fs` read, so this does work - but it
pulls a 3.9 MB generated artifact into the editor's type-check program to read one
prototype. Issue #366 step 3 also asked for a spec modelled on
`tests/railgun-turret-diagonal.spec.ts`, which is where the railgun fix's equivalent
lives.

Second: "2.1 alternate connections are defined at north-east, then rotate in quarter
turns" is an inference rather than something fire.lua states. It gives one alternate
pair and no rotation convention for the other three diagonals. The test asserts the
same model it assumes, so it cannot catch a wrong one. Nothing reaches this path
with 2.0 data, so being wrong costs nothing today - but it is worth a line on the
#187 migration checklist to confirm against regenerated data.

wormeyman (changes requested):
Reviewed at `6f82bb77` against a clean base at `03650118`.

The core of this is right, and I checked the parts that are easiest to get wrong
against the real 2.1.14 data rather than taking the PR body's word for them:

- The eight animation keys and their order match what
  `fireutil.flamethrower_turret_extension` returns in
  `base/prototypes/entity/fire.lua` (lines 206-213), and
  `graphics_set.base_visualisation.animation` is keyed the same way.
- The alternate connection values match fire.lua lines 528-529 exactly:
  `alt_direction = west (12)` with `alt_position = {-1, 0}`, and
  `alt_direction = south (8)` with `alt_position = {0, 1}`.

The test earns its place too. With both source files reverted it fails, and
inverting the `alt_position` preference fails on a 0.5 tile shift, so the cover
geometry is genuinely pinned rather than just described.

Two things to fix before this merges. The first is a measured regression on
today's committed data, and the second is a gap in what the test proves. The
other two comments are smaller and do not block. Details inline.

wormeyman:
I pushed two commits onto this branch to close out the review, since `maintainerCanModify` is on:

- `17ab6905` takes the alternate-connection branch only when the prototype declares `alt_position` or `alt_direction`, indexes `draw_fluid_turret` through `ROTATED_8WAY_KEYS[(dir >> 1) & 7]`, and lets `draw_railgun_turret` unwrap `base_visualisation` through `baseVisualisationLayers` so the two turret draws have the same shape.
- `b53aafe9` moves the test to `tests/fluid-turret.test.ts`, where it reads `data.json` off disk instead of importing it into the editor's type-check program, and adds the neighbour controls.

The first finding turned out to be wider than one entity. Drawing all 155 entities at eight directions with fluid inputs on, five went from the placeholder to a silent cardinal render at every diagonal on `6f82bb77`:

| entity | base `03650118` | `6f82bb77` | with the gate |
| --- | --- | --- | --- |
| electromagnetic-plant | FAILED | 8 layers | FAILED |
| pumpjack | FAILED | 4 layers | FAILED |
| steam-engine | FAILED | 4 layers | FAILED |
| steam-turbine | FAILED | 4 layers | FAILED |
| storage-tank | FAILED | 6 layers | FAILED |

`electromagnetic-plant` only reaches the cover code with a fluid recipe. Without one, a crafting machine has no fluid box to cover and nothing throws. With the gate in place the sweep matches base on all 155 entities at all eight directions, and every turret's output is byte-identical to base, railgun included.

Mutation checks on the new test, each reverted afterwards:

- direction compare replaced with `true`: 2 tests fail
- position compare dropped: 1 fails
- gate reverted to the bare direction test: 1 fails, on `storage-tank` at direction 2 rendering 6 layers
- old `dir / 2` indexing restored: 4 fail

The note that the north-east-then-rotate convention is an inference is now in the test header. `vp check` is clean, the 307 unit tests pass, and the railgun, sprite-data and sprite-generation Playwright specs pass against this head.

Issue #362, opened 2026-09-05
Cargo bays never draw their hatch, and four sub-visuals go missing


wormeyman:
Noticed while reviewing #356, which fixed the same shape of bug on the
electromagnetic plant: a sub-visual the draw function never reaches, so the
entity renders with a piece missing.

#356 does not touch cargo bays. `cargo-bay` is its own prototype type and goes
to `draw_cargo_bay`, a separate path from `draw_assembling_machine`, which #356
additionally scopes to `electromagnetic-plant` by name. But the class of defect
does apply here, in four places.

All line numbers are `packages/editor/src/core/spriteDataBuilder.ts` at
`e428a8e4`. None of this is measured against a screenshot yet - it is read off
the prototype data and the draw code.

## 1. The hatch is never drawn

`data.json` gives `cargo-bay` a `hatch_definitions` list whose single entry
carries `hatch_graphics` (alongside `offset`, `travel_height`,
`pod_shadow_offset` and the sounds).

`draw_cargo_bay` (1336) reads only `graphics_set.picture` and
`graphics_set.connections`. Nothing reads `hatch_definitions`, so the hatch
never renders. This is the closest match to the plant's missing core and the
most likely thing a user would call "weird looking".

## 2. Five connection pieces are unreachable

`graphics_set.connections` has 17 keys. `getCargoBayConnectionSprites` (1276)
calls `addConnectionSprites` with 12 of them: four walls, four outer corners,
four inner corners.

The five never referenced are `bridge_crossing`, `bridge_horizontal_narrow`,
`bridge_horizontal_wide`, `bridge_vertical_narrow` and `bridge_vertical_wide` -
the pieces that span between bays rather than edging one. A layout that needs a
bridge silently gets nothing there.

## 3. `render_layer` is discarded

Line 1309 strips it:

```ts
const { render_layer: _render_layer, ...spriteData } = rendition
```

That is the field saying where the piece belongs in the draw order. With it
gone, the result is composed positionally as `[...connections, ...base]` (1338),
so every connection sprite lands under the base picture whatever the game
intended. Worth checking against a real cargo bay before changing, since the
current order may be right by accident for the common case.

## 4. Only the first variant is ever used

Line 1302 takes `variants[0]` unconditionally, so the random variations the
prototype offers never appear. Cosmetic, and the least urgent of the four.

## Not a bug

The 4x4 assumption in the comment at 1283 is right: `collision_box` is
[...]

PR #379, opened 2026-09-06
Draw the cargo bay bridges, on the seam between two entities

### LINKED ISSUE #378: Cargo bay connection sprites are all drawn at the entity centre, so bays never join up
Found while measuring #362. This is a separate defect from the missing hatch and is not touched by #376.

**Two cargo bays placed side by side render as two islands with a gap between them.** In game they join into one continuous structure. The editor never draws the join.

## What is actually wrong

`getCargoBayConnectionSprites` in `packages/editor/src/core/spriteDataBuilder.ts` picks connection pieces correctly and then draws every one of them **at the entity centre**.

Each connection sprite's own `shift` is tiny - the first layer of every one of the 17 keys is `util.by_pixel(2.5, -1.0)`, so `[0.078125, -0.03125]` - because the shift is meant to position the piece *within its own tile*, not within the entity. With no per-position offset added, all of them stack on the same spot.

Measured opaque extents of all 17 keys, against a bay footprint of x -2..2, y -2..2:

| key | x | y |
| --- | --- | --- |
| `top_wall` | -1.05..1.05 | -1.14..-0.41 |
| `bottom_wall` | -1.03..1.09 | -1.41..1.19 |
| `left_wall` | -1.19..0.00 | -1.55..1.80 |
| `right_wall` | -0.02..1.13 | -1.58..1.94 |
| `top_left_outer_corner` | -1.08..1.03 | -0.98..2.05 |
| `bridge_horizontal_wide` | -1.66..1.80 | -1.98..0.50 |

Nothing reaches the bay's edge, and the four walls sit on top of each other rather than on four different sides. Every piece, bridges included, stays inside roughly x -1.7..1.8.

## The selection half already works

For two bays touching east-west, the west bay correctly drops `right_wall` and the east drops `left_wall` - 29 and 31 layers instead of the 37 a lone bay emits. The code knows they are adjacent. Only the placement is wrong, so nothing bridges the seam.

## Why: it is a per-tile system driven per-entity

Factorio 2.1 restructured `CargoBayConnections` and documents the mechanism plainly. The 12 named wall and corner keys became `tileset` plus `tileset_mapping`:

> A mapping from a bitmask index to a tileset index. A bitmask index mapped to 0 won't be drawn. Tile bitmask uses 8 bits. Bits are assigned from the top-left corner and going clockwise (top-left tile has bit 0 and right tile has bit 7).

So each **tile** (or 2x2 build-grid cell - `build_grid_size` is 2 and a bay is 4x4) computes its own 8-neighbour mask, picks its own piece, and draws it at its own offset. The editor computes one mask for the whole entity and stacks the results at the centre.

Our `data.json` is 2.0-shaped, so it still has the 12 named keys, but the roles are the same.

## Partial confirmation

A cheap test - drawing only the four outer corners, one per 2x2 cell at (±1, ±1), and nothing else - is visibly much closer to the in-game bay than what ships today. It is 19 layers instead of 37. That is consistent with "an isolated bay is four corner cells and no walls at all", but it is a hypothesis, not the derived rule. The exact `tileset_mapping` translation still needs working out.

Reference for what it should look like: <https://wiki.factorio.com/File:Cargo_bay_entity.png>

## Scope

- 330 cargo bays in the committed corpus, plus 13 cargo landing pads. The commonest arrangements are neighbours on three sides (95 bays) and the opposite three (94), with 76 fully surrounded - so almost every corpus bay is one that should be joining to something.
- `cargo-landing-pad` goes through the same helper and has the same defect.
- `space-platform-hub` has a `connections` entry but `draw_space_platform_hub` never calls the helper at all, and `isCargoBayLike` does not match a hub, so a bay placed against one draws an exterior wall on the shared edge. Both of those are already noted in #364 and probably want fixing in the same change.

## Knock-on for #362

This reframes item 2 of #362, the five unreachable `bridge_*` pieces. That issue is [refuted](https://github.com/FactoryGameFan/factorio-blueprint-editor/issues/362#issuecomment-5556859605) on the grounds that bridge choice is not a function of the 8-neighbour mask the editor computes - which is true, but the deeper reason is that the whole system is being driven at the wrong granularity. Bridges are worth revisiting only after the per-tile placement is right.

## Testing

`tests/__fixtures__/sprite-data.json` will move in all five sections for `cargo-bay` and `cargo-landing-pad`. Layer counts will change, not just hashes, since the number of pieces per bay changes. A green fixture says nothing about whether bays *look* joined, so this one needs a rendered check against the wiki shot, not just a re-record.

wormeyman:
Finishes the per-cell rewrite and draws the five `bridge_*` connection pieces, so cargo bays, landing pads and platform hubs render as one structure instead of pieces with holes between them.

Closes #378. Closes item 2 of #362.

## The rule, and where it comes from

A bridge sits on the shared edge **between** two entities, not on a cell of either. That is why no cell mask selects one, and why Factorio 2.1 leaves all five keys outside `tileset_mapping`. I read that silence as "unreachable" in #362 and it was wrong.

`tools/oracle/probe-cargo-bay-render.mjs` now renders twelve arrangements in Factorio 2.0.77 and captures each twice, with and without the entities, so differencing gives an exact mask. Each candidate piece is scored over the pixels it would itself cover, which is the only instrument that separates them - the four wide and crossing frames share a base plate, so their silhouettes are near identical and a coverage score cannot tell them apart.

At every seam measured, exactly one of the eight candidates beats drawing nothing and the other seven make it worse. Gain is the drop in mean absolute RGB error over the piece's own pixels:

| adjacency | winner | gain | runner-up |
| --- | --- | --- | --- |
| side by side, 4-tile edge | `bridge_horizontal_wide` | +7.8 | -1.6 |
| stacked, 4-tile edge | `bridge_vertical_wide` | +10.4 | -3.9 |
| side by side, 2-tile edge | `bridge_horizontal_narrow` | +1.4 | -1.9 |
| stacked, 2-tile edge | `bridge_vertical_narrow` | +11.7 | -4.8 |
| four bays meeting | `bridge_crossing` | +20.6 | +2.1 |
| a pad and two bays meeting | `bridge_crossing` | +7.5 | -2.3 |

So:

- The key names the direction the bridge **spans**. Entities side by side take `bridge_horizontal_*`, stacked ones take `bridge_vertical_*`.
- A 4-tile shared edge takes `_wide`, a 2-tile one `_narrow`. A longer edge is split into 4-tile spans, which is what a landing pad's 8-tile side does anyway when two bays sit against it.
- A cell corner takes `bridge_crossing` when all four quadrants are covered **and** the covering entity changes both left to right and top to bottom. Both halves are load bearing: without the second, every ordinary seam midpoint would take a crossing on top of the bridge already there, because two entities either side of a seam do make its midpoint a four-quadrant point.
- Each piece is drawn once, by the entity west or north of the seam and by the one north-west of a crossing. The neighbour is redrawn with it because all three names are in `EntityContainer`'s update group.
- Nothing spans a gap. Two bays 2 tiles apart draw no join at all in game. That capture is the control, and it is why "bridge" does not mean what the word suggests.

Three quadrants is not enough, and neither is four quadrants covered by one entity - both measured, both scoring worse than drawing nothing.

## Result

Share of the game's entity pixels wrong, shadows counted on both sides:

| arrangement | before this branch | after |
| --- | --- | --- |
| two side by side | 12.6% | 9.1% |
| stacked pair | 15.1% | 10.1% |
| a row of three | 15.6% | 9.0% |
| a 2x2 block | 18.4% | 7.6% |
| an L | 12.3% | 5.9% |
| six in a 2x3 block | 23.2% | 12.4% |

Against what ships on the base branch the numbers are 29% to 32% wrong, because every piece was drawn at the entity centre.

## Three defects found on the way, none of them the bridges

**A landing pad and a platform hub drew no connection sprites at all.** `getCargoBayConnectionSprites` derived its footprint from a hard-coded 4x4, so an 8x8 entity sampled tiles inside itself, always looked fully surrounded, and every guard fell through. All 13 corpus landing pads produced one identical digest; they now produce four, and a lone pad goes from 17 layers to 72. The hub's 55 connection layers match the pad's exactly, which is the check that the geometry is right - two lone 8x8 entities must draw the same thing.

**The footprint origin was rounded.** Loading a blueprint re-centres it, and a blueprint whose extent is odd puts everything in it on a half tile: the all-entities blueprint at four directions puts the landing pad at x -73.5. Rounding there moved the cells half a tile off the entity while the sprites stayed put, so the same lone pad drew different sprites depending on where it had been dropped. That is what the odd fixture digests were saying.

**Only `cargo-landing-pad` declares `build_grid_size`.** `cargo-bay` and `space-platform-hub` declare none, so they snap to whole tiles and two bays can sit an odd number of tiles apart. Asking a neighbouring cell's middle whether it was occupied then read straight across a 1-tile gap, and read it differently from each side - one bay opened its wall while the other kept it, a disagreement `getSurroundingEntities` is too narrow to redraw across. Occupancy is now asked of the tile immediately across the edge.

Also: the entity number is passed into the draw data rather than looked up in the grid. The grid keeps overlapping entities in one cell and hands back the smallest, so an overlapping import could have made a landing pad believe it was the bay sitting inside it, and then both would have drawn the same crossing.

## The failing assertion at 4c8ec5e6 was the assertion

A layer count is not a monotone function of neighbours, so it cannot express "this bay joined to something". A bay that gains a west neighbour swaps its two left outer corners, 4 + 5 layers, for a top and a bottom wall, also 4 + 5. Measured, a bay covered on its west or north stays at 26 while one covered on its east or south drops to 25. The spec compares digests now, and the header comment that claimed the count falls with every neighbour is gone.

## Testing

- `tests/cargo-bay-connections.test.ts` - 24 pure tests. The cell rule against the game's 175-mask table, and the bridge and crossing rules against the measurements above. Mutation checked: dropping the crossing's second guard fails 5, always taking the wide piece fails 2, letting both entities of a pair draw the seam fails 1.
- `tests/cargo-bay-connections.spec.ts` - placement in the real editor, including the 1-tile-gap regression. That one fails if the sampling goes back to a neighbouring cell's middle.
- `tests/__fixtures__/sprite-data.json` re-recorded. Only 2 of its 5 sections move, not all 5 as #378 predicted: the other three run without a position grid, where the helper returns an empty array by construction and nothing can change.
- `vp check` clean, 324 unit tests, 265/265 Playwright.

## Left open

- `render_layer` is still discarded, #362 item 3. It now costs 3.5% to 7.1% of the entity's pixels, up from the 684 measured before, because one entity draws a bridge over its neighbour. The local sort that issue suggests cannot fix the grown half - bridge versus neighbour order is decided by `EntitySprite`'s global sort on `entityPos.y` - so it wants its own change.
- `variants[0]` is still taken unconditionally, #362 item 4. Measured on the bridges: which variant the game uses differs from seam to seam, so matching it needs a position hash we would be inventing.
- Two shapes in the rule are generalisations rather than measurements, and the comments say so: an odd shared edge of 1 or 3 tiles, and an 8-tile edge between two 8x8 entities. Neither is in the corpus, and the second cannot happen in game because a force gets one landing pad per surface.

https://claude.ai/code/session_01RLAG3Ha1TYeda24FssHzEV

## Summary by CodeRabbit

- **New Features**
  - Added automatic connection graphics for cargo bays, cargo landing pads, and space-platform hubs.
  - Supports horizontal and vertical bridges, diagonal joins, crossings, gaps, and offset layouts.
  - Connection rendering now accounts for entity size and fractional positions.
  - Space-platform hubs display surrounding connection sprites alongside their existing graphics.

- **Bug Fixes**
  - Updated surrounding connection graphics when supported entities are added or changed.

- **Documentation**
  - Added guidance for cargo-bay connection behavior and graphical rendering checks.


wormeyman:
Fixed in c5f1aad3. The probe now requires `done.txt`, the `placed.json` control and both screenshots for every one of the twelve cases before it copies anything, and names what is missing otherwise. Good catch - it is the repo's own rule for probes (the dump decides whether a run worked, not the exit code) and this one was only checking that the output directory existed.

Issue #415, opened 2026-09-09
[BUG] Windows CRLF checkouts fail formatting and the rail-box-orientation fixture test


byalex33:
**Describe the bug**
A fresh Windows checkout with Git's `core.autocrlf=true` fails the repository's checks without any source changes. The formatter expects LF, and the rail-box-orientation analyzer test compares its LF stdout with the fixture's checked-out CRLF bytes.

**To Reproduce**
1. On Windows, set `core.autocrlf=true` for a fresh clone of this repository and check out `wormeyman-space-age-support` at `1e19bf5526bdf58afa96989e65e467f706e91bed`.
2. Install the declared toolchain/dependencies and run `vp check` and `vp test run`.
3. `vp check` reports format failures throughout the otherwise unchanged checkout. `vp test run` fails `the committed fixture is what the analyzer produces from it` in `tools/oracle/analyze-rail-box-orientation.test.mjs`.

**Expected behavior**
The documented Windows contribution workflow should pass baseline checks regardless of the Git for Windows line-ending conversion setting, or the checkout policy should enforce the line endings required by the tools.

**Desktop**
- OS: Windows, Git for Windows with system `core.autocrlf=true`.
- Browser: not applicable; these checks are browser-free.

**Additional context**
Discovered while validating the #155 data refresh, independently confirmed against an untouched checkout of the same base commit. The full implementation-test run reported 280 formatter failures and 438/439 unit tests passing; those counts come from that run, not the standalone diagnostic below.

A standalone diagnostic on the untouched base ran the analyzer using Node and compared its output with `tools/oracle/fixtures/rail-box-orientation.json`:

```json
{"fixtureCRLF":470,"outputCRLF":0,"rawEqual":false,"normalizedEqual":true}
```

Here `normalizedEqual` replaces only the fixture's `\r\n` sequences with `\n` before comparison. The analyzer values are identical; the failure is line endings alone. `git config --show-origin --get core.autocrlf` reports `true` from Git for Windows' system configuration. This checkout has no `.gitattributes` enforcing LF.

No oracle expectations were changed or re-recorded. This is separate from Windows executable-path handling in oracle probes. Searched open and closed issues for CRLF, autocrlf, line endings, Windows formatting, and this analyzer before filing; found no existing report of this failure.

wormeyman:
Checked while testing the devcontainer in #472, since "use a container" looks like an answer to this. It only helps if the clone is made on the Linux side.

[...]

PR #422, opened 2026-09-09
Preserve blueprint parameters and stock connections

### LINKED ISSUE #336: serialize() drops blueprint parameters and stock_connections, silently
`Blueprint.serialize()` never emits top-level `parameters` or `stock_connections`, so a parametrised blueprint round-trips as a de-parametrised one with no warning.

**This is not a niche shape.** Decoding the committed corpus: **97 of 367 blueprints (26%) carry top-level `parameters`**, and 28 carry `stock_connections`. They are ElderAxe's parametrised rail stations, where parametrisation is the point of the book. Losing it turns "Select the item to load" into a hardcoded station.

**The data is gone at parse time, not serialize time.** `Blueprint`'s constructor copies five top-level fields and never touches `data.parameters`, so adding a line to the return literal would ship a no-op. This was mis-stated in #329 as a three-line fix.

Damage pattern worth knowing: `Book` keeps raw entries and spreads them through, so only the *active* blueprint loses the data - but `selectBlueprint()` saves the active one before switching, so **browsing a book strips exactly the blueprints you looked at**. A bare blueprint always loses it.

**Shape of the fix** (follow `schedules`, which is carried verbatim for the same reason):
- `types.ts` `IBlueprint`: +2 members, `unknown[]` is fine, nothing reads them
- `Blueprint.ts`: +2 fields, +2 constructor assignments beside `description`, +2 keys in the return literal

**Testing:** `tests/blueprint-round-trip.spec.ts` hashes `serialize()` output against a pinned constant, so it currently *bakes in* the loss rather than catching it. Fixing this moves `serializedHash` while every count holds - that file's header forbids regenerating it silently, so budget the justification. A round-trip unit test belongs in `Blueprint.test.ts`; add a book case too, for the strip-on-the-way-past behaviour.

Schema already accepts both (`blueprintSchema.json`), added to stop validation rejecting real blueprints rather than to preserve anything.

<sub>Split out of #329. Verified against source before filing; citations are `path:line` at the time of writing.</sub>

byalex33:
Blueprint import/export now retains top-level parameters and stock_connections, including when browsing and revisiting nested book entries. Populated arrays, empty arrays and absent fields keep their values.

Added unit and nested-book round-trip regressions. The corpus hash changes from 568918011 to 1944909146 solely because metadata is restored: stripping those two fields reproduces the previous hash, every count and position checksum holds, and 97 parameter-bearing / 28 stock-connection-bearing blueprints survive.

Validation: vp check; 443 unit tests; six Playwright corpus/book serialization tests. Local unit validation required LF normalization of the existing rail-box fixture (tracked separately in #415); no fixture changes are included.

Closes #336.

## Summary by CodeRabbit

- **Bug Fixes**
  - Blueprint metadata in `parameters` and `stock_connections` is now preserved when blueprints are loaded, serialized, and exported.
  - Metadata remains intact when working with nested books and visiting blueprint entries.
  - Blueprints without these optional fields continue to round-trip correctly.
- **Tests**
  - Added coverage for populated, empty, and absent metadata fields to help ensure consistent blueprint round-tripping.

### DIFF
--- a/packages/editor/src/core/Blueprint.test.ts
+++ b/packages/editor/src/core/Blueprint.test.ts
@@ -1,6 +1,7 @@
 import { beforeAll, describe, expect, it } from 'vite-plus/test'
 import { Blueprint } from './Blueprint'
-import { ISignal } from '../types'
+import { Book } from './Book'
+import { IBlueprint, ISignal } from '../types'
 import { encode, getAndClearLoadWarnings, getBlueprintOrBookFromSource } from './bpString'
 import { loadData } from './factorioData'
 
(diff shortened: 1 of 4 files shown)


byalex33:
@coderabbitai full review

PR #425, opened 2026-09-10
Refresh @xmldom/xmldom and fast-uri to clear 14 Dependabot alerts

wormeyman:
Both packages were already inside their declared semver ranges, so this is a lockfile refresh with no manifest change.

| package | from | to | pulled in by | alerts cleared |
| --- | --- | --- | --- | ---: |
| `@xmldom/xmldom` | 0.8.13 | 0.8.15 | `pixi.js ^0.8.13` | 10 |
| `fast-uri` | 3.1.5 | 3.1.7 | `ajv ^3.0.1` | 4 |

All 14 are runtime scope. 0.8.15 covers both patch levels the advisories ask for - 8 need 0.8.15 and 2 need 0.8.14.

The diff touches 4 version lines in `package-lock.json` and nothing else.

## What this does not fix

The 15th alert is `sharp` 0.35.2 (high, dev scope), reached through `miniflare`, which pins it exactly. It needs `wrangler >= 4.131.0` - that version is ~7 hours old at time of writing, so it is left as a separate decision rather than folded in here.

## Verification

- `npm audit`: the `@xmldom/xmldom` and `fast-uri` advisories are gone; only the `sharp` chain remains
- `npm run test:unit`: 22 files, 249 tests, all passing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3006,9 +3006,9 @@
             "license": "BSD-3-Clause"
         },
         "node_modules/@xmldom/xmldom": {
-            "version": "0.8.13",
-            "resolved": "https://registry.npmjs.org/@xmldom/xmldom/-/xmldom-0.8.13.tgz",
-            "integrity": "sha512-KRYzxepc14G/CEpEGc3Yn+JKaAeT63smlDr+vjB8jRfgTBBI9wRj/nkQEO+ucV8p8I9bfKLWp37uHgFrbntPvw==",
+            "version": "0.8.15",
+            "resolved": "https://registry.npmjs.org/@xmldom/xmldom/-/xmldom-0.8.15.tgz",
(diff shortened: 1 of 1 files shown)


wormeyman:
Reviewed. **Correct on every axis I checked**, and good to merge.

- All integrity hashes match the registry.
- Both bumps sit inside their parents' declared ranges, so this is a lockfile-only change. `package.json` is untouched and the `vite-plus` alias is intact.
- The "14" in the title is exactly right: 10 xmldom alerts plus 4 fast-uri.

**#426 is a strict subset of this.** Its xmldom hunks are byte-identical, and merging both gives a tree identical to this PR alone - measured by tree hash rather than by reading file lists:

```
merge-tree(pr425, pr426) = 10a895b4b1666e13997da04843c6aa8d04426b45
tree(pr425)              = 10a895b4b1666e13997da04843c6aa8d04426b45
```

So merging this closes #426 with nothing left over.

**One finding against the repo, not against this PR.** The `sharp` advisory `GHSA-rgj7-g3m4-5g8c` survives - it needs `sharp >= 0.35.4`, which arrives through Wrangler. So the open alert count after merging this is 1, not 0. The commit body already says so; flagging it only because the count dropping is easy to read as "clean". #428 is the PR that clears it.

Issue #420, opened 2026-09-09
[BUG] Suspected intermittent CI hang in HTML blueprint-source rejection test


byalex33:
**Describe the bug**
One CI execution of `tests/blueprint-sources.spec.ts:268`, `a 200 that is really an HTML error page fails as a corrupt blueprint`, timed out after 120000ms. This is an observed intermittent test hang with an unknown cause, not a confirmed editor or codec defect.

**To Reproduce**
- Observed on PR #418, head `1f1dc574`, CI run https://github.com/FactoryGameFan/factorio-blueprint-editor/actions/runs/34395552535, Playwright shard 1/4.
- Run the committed browser suite against the development servers. The test intercepts `/corsproxy`, responds HTTP200 with `<!DOCTYPE html><html><body>Sign in to continue</body></html>`, awaits the source-import rejection, and removes the route.
- This execution failed with `Test timeout of 120000ms exceeded.` The shard had 87 passes and this single failure; the other three shards passed.
- Reproduction is incomplete: the exact unchanged test passed 10/10 locally in 14.6 seconds, with individual tests around 0.8–1.1 seconds.

**Expected behavior**
The intercepted HTML response should promptly reject as an invalid blueprint, and the test should finish without weakening its assertions or raising its timeout.

**Desktop**
- Failure: GitHub Actions Linux Chromium browser test job.
- Local diagnostic: Windows, Node24.20.0, Vite+0.3.0, Chromium, same PR source with isolated sprite server.

**Additional context**
The uploaded `playwright-report-shard-1` artifact includes `error-context.md` showing an initialized editor page and the 120-second timeout, plus `localpreview.log`. It does not include a call stack or execution trace locating the stalled operation.

The exact HTML fixture normalizes to 45 base64 characters. `atob` rejects this synchronously before `pako.inflate` is called; `decode` catches and rejects that error. This rules out a pako inflate hang for this exact body but does not establish whether browser transport, route handling, teardown, or something else caused the CI timeout.

Local check, with traces enabled:

```shell
playwright test tests/blueprint-sources.spec.ts --grep 'a 200 that is really' --repeat-each 10 --trace on
```

No source changes, timeout changes, assertion changes, or automatic test retries were added. One failed-job-only CI retry is underway to distinguish an isolated infrastructure event from a repeating failure. The cause remains unknown; this issue preserves the evidence rather than treating the passing local repetitions as proof that no bug exists.

[...]

PR #428, opened 2026-09-10
Fix advisory, checkbox, Windows checkout and generated-data CI defects

### LINKED ISSUE #415: [BUG] Windows CRLF checkouts fail formatting and the rail-box-orientation fixture test

### LINKED ISSUE #416: [BUG] Shared oracle runner resolves read-data to nonexistent bin/data on Windows
**Describe the bug**
`tools/oracle/factorio-probe.mjs` writes `read-data=__PATH__executable__/../data`. On Windows the executable lives in `Factorio/bin/x64`, so this resolves to nonexistent `Factorio/bin/data` rather than `Factorio/data`.

**To Reproduce**
1. Set `FACTORIO_BIN` to an installed Windows `bin/x64/factorio.exe`.
2. Run a probe using `runProbe`, for example `node tools/oracle/probe-entity-tile-size.mjs`.
3. The game stops before loading mods. While probing #183 with the unchanged runner, it reported `Error configuring paths: There is no package core in .../Factorio/bin/data. Deduced executable directory: .../Factorio/bin/x64, read data: __PATH__executable__/../data`.

**Expected behavior**
Find the installed data directory on supported executable layouts, or accept an explicit data-directory override.

**Desktop**
Windows 11; Factorio 2.0.77 (build 84539, win64, Steam, Space Age); Node 22.17.1; repo commit `1e19bf55`.

**Additional context**
Changing only the isolated temporary config to the actual installation data directory let the #183 probe initialise and produce its dump. No user profile was changed. The current relative path fits the macOS layout used originally. Found incidentally while testing #183; searched open and closed issues for oracle Windows, read-data, and factorio-probe before filing.

### LINKED ISSUE #419: Generated dataset-only PRs skip web checks and Playwright
**Describe the bug**
The changed-area classifier treats every path beneath `packages/exporter/` as Rust-only, including committed `data/output/data.json` and sprites that the web editor consumes. A generated-data-only pull request skips `checks` and every Playwright shard, so browser regressions in the refreshed dataset are not tested before merge.

**To Reproduce**
1. Open a pull request against `wormeyman-space-age-support` with only `packages/exporter/data/output/data.json` changed (an exporter README change does not alter the result).
2. Inspect the CI `Detect changed areas` job.
3. It selects `rust=true`, `web=false`, and skips web checks/Playwright.

Observed on PR #418 at `0093d1d26d93e982656c5b460eac2757abeae677`: https://github.com/FactoryGameFan/factorio-blueprint-editor/actions/runs/34395174454 . The PR changes exactly the exporter README and generated JSON. This is the initial run; adding a non-exporter file later may incidentally turn the web gate on without fixing the classifier.

**Expected behavior**
Changes to committed exporter output should run the web checks and browser suites. Source-only Rust changes can retain their narrower gate.

**Desktop**
GitHub Actions CI on base `1e19bf55`; independent of browser/desktop OS.

**Additional context**
`.github/workflows/ci.yml` currently uses `grep --quiet --invert-match '^packages/exporter/'` to determine `web`. The application loads the generated JSON and textures directly, and the exporter review checklist explicitly requires dataset-sensitive browser coverage. Those 11 tests passed locally for PR #418, but the automated pre-merge gate omitted them. Found while testing #155; searched open/closed issues before filing.

### LINKED ISSUE #423: [BUG] Checkbox ignores a second click before the next render
**Describe the bug**
A second click on the Snap to grid checkbox can be ignored when it arrives before Pixi renders the graphics rebuilt by the first click. Found while running unrelated icon-dialog coverage for #421. One ordinary run failed the existing untick/re-tick regression; five focused retries passed. A paused-render probe reproduces the lost click deterministically.

**To Reproduce**
1. Load a blueprint with snap-to-grid {x:20,y:18} and open Blueprint Info.
2. After the dialog has rendered, pause the Pixi ticker (`G.app.ticker.stop()`) to hold the gap between frames.
3. Click Snap to grid: `G.bp.snapToGrid` becomes undefined as expected.
4. Click the same point again before rendering: it remains undefined.
5. Call `G.app.render()` and click once more: it becomes {x:20,y:18}.

The ordinary affected test is `tests/blueprint-info-editor.spec.ts`, “unticking and re-ticking Snap to grid keeps the grid size instead of replacing it with 1x1”. Observed at its final assertion: expected {x:20,y:18}, received undefined.

**Expected behavior**
Two successive clicks toggle the checkbox off and back on, even within a frame.

**Desktop**
Windows, Chromium via Playwright 1.62, Node 22.17.1, commit 8973b4ab based on 8635b687.

**Additional context**
Suspected mechanism: ToggleControl.checked destroys and replaces its Graphics children on each change. The replacement graphics appear to lack usable hit-test geometry until the next render. This is an input/render timing diagnosis, not a grid-size persistence failure. Holding the render gap makes the behavior deterministic; normal user reproduction remains intermittent and frame-rate dependent.

The icon PR changes only the named blueprint icon fallback; this synthetic blueprint has no explicit icons and the affected checkbox path is unchanged. No retries or sleeps were added to the production test.

### LINKED ISSUE #424: [BUG] Locked transitive dependencies now report xmldom, fast-uri, and sharp advisories
**Describe the bug**
A fresh locked installation at 8635b687305eb13ce05700901c73d4c100bd99c5 reports five high-severity vulnerable package entries from three transitive dependency families. This was encountered while setting up validation for the open PRs; it predates their changes. This report establishes the installed versions and npm audit result, not an exploitable application path.

**To Reproduce**
1. Check out commit 8635b687305eb13ce05700901c73d4c100bd99c5.
2. Use the pinned Vite+ 0.3.0 runtime and run `vp install --frozen-lockfile`.
3. Run `vp pm audit --json`.

**Expected behavior**
Resolve actionable transitive advisories or document why a parent pin prevents the update and what application exposure exists.

**Actual behavior**
- `@xmldom/xmldom` 0.8.13: several XML parsing/serialization advisories, including https://github.com/advisories/GHSA-w2rr-34g9-rvrj and https://github.com/advisories/GHSA-8344-3jmq-59r6. `pixi.js` accepts `^0.8.13`, so its parent range allows a patched release.
- `fast-uri` 3.1.5: https://github.com/advisories/GHSA-5jgf-p345-68v8 and related canonicalization advisories. Audit gives the fixed boundary as 3.1.6. `ajv` accepts `^3.0.1`.
- `sharp` 0.35.2: https://github.com/advisories/GHSA-rgj7-g3m4-5g8c, with fixed boundary 0.35.4. `miniflare` pins **exactly** 0.35.2. Audit also counts the dependent Miniflare and Wrangler entries, giving five high-severity entries overall. Upgrade the compatible Wrangler/Miniflare stack when available; do not independently override sharp or apply the audit suggestion to downgrade Wrangler to 4.15.2.

**Desktop**
Windows, PowerShell, managed Node 24.20.0, Vite+ 0.3.0; observed 2026-09-09 UTC. Browser not applicable to the audit reproduction.

**Additional context**
Open and closed issues were searched for xmldom, fast-uri, sharp, and GHSA; no matching advisory report was found. Issue #158 describes an earlier zero-audit baseline but does not track these advisories. No proof of exploitability in the deployed editor has been established, and no unrelated dependency changes were folded into the PR review work.

byalex33:
Fixes five reported defects and adds failure diagnostics for the still-unexplained HTML import timeout:

- Resolve xmldom and fast-uri advisories and upgrade the compatible Wrangler/Miniflare/workerd stack to obtain patched sharp. npm audit now reports zero vulnerabilities. This includes the same xmldom/fast-uri refresh as #425 and extends it to sharp; #425 and #426 are left unchanged.
- Keep checkbox/radio hit areas stable when graphics are replaced between renders. A paused-clock regression reproduces the lost second click with the fix removed and passes with it restored.
- Run web checks and Playwright for generated exporter JSON and sprites, with a regression exercising the actual CI predicate.
- Find oracle read-data in macOS and Windows/Linux installations across all three config writers.
- Enforce LF text checkouts. Four previously committed CRLF text files are normalized without content changes; binary assets remain unchanged.
- Retain blueprint-source failure traces for #420. The HTML rejection passes locally; no root cause was established, so this does not claim to fix or close that issue.

Validation: vp check passed; all 445 unit tests passed; seven CI classifier cases passed; website build and Worker HTTP smoke passed; built-in browser verified snapping preserves a 20x18 grid and production editor loads. The real Windows Factorio 2.0.77 entity-tile-size probe completed successfully using an isolated temporary profile, without recapturing fixtures. All 345 Playwright tests passed locally (7.7 minutes).

Closes #424
Closes #423
Closes #419
Closes #416
Closes #415
Refs #420


wormeyman:
Thanks for this - six real defects in one pass, and the dependency half is genuinely clean work: 37 of 37 integrity hashes verify, the wrangler/miniflare/workerd/undici stack moves as one unit exactly as CLAUDE.md requires, all 15 open Dependabot alerts clear, and the lockfile reproduces byte-identically from `npm install --package-lock-only` plus `npm update`, so it really is npm's output.

Two things to change before merge, one timing ask, and some non-blocking follow-ups.

---

## Blocker 1 - port the CI self-test to `.test.mjs`

`scripts/ci-changes.test.sh` has the right instinct: it runs the *live* predicate out of `ci.yml` instead of copying its path rules, and I confirmed it catches the `echo | grep` SIGPIPE regression the workflow comment warns about. Three problems, and moving it to `scripts/ci-changes.test.mjs` fixes all three at once.

1. **It fails silently on a reworded predicate.** Line 6 anchors the `sed` range on the literal text `if grep --quiet --invert-match`. A behaviour-preserving rewrite of that line makes the extraction empty, and line 7's bare `test -n "$classifier"` then exits 1 printing nothing at all.
2. **A failure there hides the whole unit suite.** It shares a `run:` block with `vp test` (`ci.yml:126-128`) under `bash -e`, in a step called "Unit tests". So the author sees `Error: Process completed with exit code 1` under "Unit tests" and no unit test results.
3. **`vp test` never runs it locally.** `vite.config.ts:266` includes `scripts/**/*.test.mjs`, not `.sh`. `scripts/localpreview.test.mjs` is the existing sibling in that directory, so `.test.mjs` is already the convention.

If you would rather keep it as shell, at minimum give line 7 a message:

```sh
test -n "$classifier" || { echo "could not extract the web predicate from ci.yml" >&2; exit 1; }
```

## Blocker 2 - `QuickbarPanel` has the same defect as #423, and **the fix here does not transfer to it**

`QuickbarPanel.changeActiveQuickbar` (`QuickbarPanel.ts:159-166`) does `slotsContainer.removeChildren()` then `generateSlots(...)`, building brand new slots, from a `pointerdown` handler (`:63`) and from the `X` keybind (`Editor.ts:915-923`). Same shape as the checkbox.

I reproduced it. #423 held the frame gap open with `G.app.ticker.stop()`; `G` is not on `window`, so I dispatched the `X` keypress and the slot click inside one synchronous `evaluate` block instead, which makes a render between them impossible. The observable is that clicking an empty slot opens the Inventory dialog, which `__fbe_test.openDialogCount()` can see:

```
AIM CHECK  real click on slot 0        -> openDialogCount = 1   (coords on target)
CONTROL    synthetic click, no X       -> openDialogCount = 1   (synthetic events work)
PROBE      X + click, same frame       -> openDialogCount = 0   <-- click swallowed
RECOVERY   click after frames passed   -> openDialogCount = 1
```

**The important part:** I then transcribed this PR's own fix onto `Button` - `this.hitArea = new Rectangle(0, 0, width, height)` in its constructor, which every slot inherits - and re-ran. **It does not fix it.** The probe still reads 0. (I proved the patch was really being served before believing that: widening the hit area to `(-200, -200, width+400, height+400)` visibly changed which clicks landed, so the instrument was live.)

The reason is worth writing into the code, because it also corrects the comment at `ToggleControl.ts:56-58`. The mechanism is not "replacement cached graphics have no rendered hit geometry" - `GraphicsContext.containsPoint` tests the recorded draw calls and needs no render at all. It is that a freshly created container's **world transform** is only written by `updateRenderGroupTransforms`, which runs inside `render()`. Between frames the hit test inverts a stale matrix and misses. `hitArea` works for `ToggleControl` because the *control itself* survives the rebuild and its transform is already correct - only its children are replaced. In `QuickbarPanel` the whole slot is new, so there is no correct transform for `hitArea` to be measured against.

What does fix it is not rebuilding at all. Reassigning the existing slots takes the probe to `openDialogCount = 1`:

```ts
public readonly changeActiveQuickbar = (): void => {
    let itemNames = this.serialize()
    // Left shift array by 10
    itemNames = itemNames.concat(itemNames.splice(0, 10))
    this.slots.forEach((slot, i) => {
        const name = itemNames[i]
        if (name) {
            slot.assignItem(name)
        } else {
            slot.unassignItem()
        }
    })
}
```

Verified: with that in place the probe reads `PROBE ... -> openDialogCount = 1`.

## Timing - please regenerate the lockfile once the floor clears

The lockfile was written with the repo's own supply-chain floor bypassed. Root `.npmrc` sets `min-release-age=1` (24 hours), and regenerating under it fails today:

```
npm error notarget No matching version found for wrangler@^4.131.0 with a date before 9/9/2026, 8:05:23 PM.
```

`wrangler@4.131.0` was published `2026-09-10T15:19:06Z` and `miniflare@5.20260910.0-alpha` at `15:16:26Z`, so both clear the 24-hour floor **today at 15:19Z**. CI is unaffected either way, because `vp install --frozen-lockfile` installs pinned versions rather than selecting them - the point is only that the lockfile should be reproducible from its own manifests under the committed config.

Cost of waiting is zero: the only advisory forcing the fresh wrangler is the `sharp` one, and `sharp` is `"dev": true`, reachable solely through `miniflare`.

---

## Non-blocking, for follow-up issues rather than this PR

- **`Closes #415` overclaims.** The sweep itself is complete - zero tracked blobs still carry CRLF, all four normalized files are byte-identical after CR removal, and all 3,073 binaries are correctly classified `-text`. But git rewrites a working-tree file only when its blob changes, so the contributor who filed #415 keeps 332 stale CRLF files, `git status` stays clean, and `git add --renormalize .` stages nothing. The real recovery is deleting the text files and checking them out again, and nothing documents that. Suggest dropping the `Closes` to a `Refs`.
- **Oracle `FACTORIO_BIN`.** `dirname(bin)` with no `realpathSync` means a bare or symlinked `FACTORIO_BIN` no longer resolves, and where a sibling directory holds a `core/` it silently reads a *different* install. The PR's own test diff is the evidence - it had to change `bin: 'factorio'` to an absolute path. `tools/oracle/README.md` is not updated either.
- **Trace diagnostics cover 1 spec file of 65.** The upload step is correct and `retain-on-failure` is the right mode for 1.63.0, but `playwright.config.ts` sets no `trace`, so the other 64 specs still record nothing. The same one line in the config would cover them all.
- **`ToggleControl.ts:56-58` comment** - see the mechanism above; "cached graphics have no hit geometry" names the wrong cause. The `.clone()` on the other hand is load-bearing and not cargo cult: `Bounds.rectangle` returns one cached object that later `.width` reads would rewrite in place.

The `ToggleControl` one-liner itself is a correct fix for a real defect, and the new spec does fail without it.

wormeyman:
The dependency half of this PR has landed separately. #425 refreshed `@xmldom/xmldom` and `fast-uri`, and #440 moved wrangler to 4.131.1, which brings sharp 0.35.4. `npm audit` is at 0 on the base branch and Dependabot has no open alerts, so I've closed #424.

That half is also the only part that conflicts. Against the current base, `git merge-tree` reports conflicts in `package-lock.json` and `packages/worker/package.json` and nowhere else. Could you reset those two files to the base versions and remove `Closes #424` from the description? The rest should then merge cleanly, and my earlier note about regenerating the lockfile no longer applies.

The other five fixes aren't on the base branch yet, so this PR is still where they belong. The two blockers from my review above still apply:

1. Move `scripts/ci-changes.test.sh` to `scripts/ci-changes.test.mjs` so `vp test` runs it.
2. Fix `QuickbarPanel.changeActiveQuickbar` too. The `hitArea` change doesn't carry over there, but reassigning the existing slots instead of rebuilding them does.

The non-blocking notes still stand as well, mainly changing `Closes #415` to `Refs #415`.

wormeyman:
@coderabbitai review
@claude review

wormeyman:
All three review findings are addressed in `6db6d23f`, with one correction to the suggested fix.

**The classifier test blocking `vp test` (Major).** Confirmed: GitHub runs a step with `bash -e`, so a failing `bash scripts/ci-changes.test.sh` exited before `vp test` and the run reported nothing about the 569 unit tests that never ran. It is now `scripts/ci-changes.test.mjs` and `vp test` discovers it through the `scripts/**/*.test.mjs` glob `vite.config.ts` already carries for `localpreview.test.mjs`. The separate shell invocation is gone, so a failure there is one test among the rest, and the check now runs for anyone running `vp test` locally, which the shell script never did.

**The silent extraction failure (Minor).** Also fixed, and one step further. Both boundary markers are asserted by name with the marker in the message, and a separate case asserts the extracted text still contains an `if grep`, a `web=true` and a `web=false`. An extraction that matches nothing now fails loudly instead of leaving every case passing against empty input. Shellcheck's SC2016 does not follow, because there is no shell left in the file.

**The symlinked `FACTORIO_BIN` (Minor).** Real, and fixed. One correction worth recording: the suggested diff imports `realpathSync` from `node:path`, which has no such export. It is in `node:fs`, so the patch as written would have thrown at import. The committed version resolves the executable first and deliberately leaves a path that does not exist alone, so the existing error still names the binary that was asked for rather than failing inside `realpathSync`.

A note on the new test, because the reason is not obvious from reading it: on macOS `realpathSync` also resolves `/var` to `/private/var`, and `tmpdir()` sits behind that link, so the expected path has to be resolved the same way or it asserts against a prefix the runner correctly no longer writes.

Both fixes are mutation-checked against the test that should catch them. Removing the symlink resolution fails only the new symlink case, leaving the other four green. Reverting the `data/output` clause in `ci.yml` fails all four dataset cases and leaves all four controls green. `vp check` is clean and `vp test` is 569 passing across 42 files.

wormeyman:
@byalex33 this branch now has commits on it that you did not write. Eric asked me to bring it up to date and review it, and I pushed the result to your branch through maintainer access. Summary below, so four commits you did not write are not the first thing you meet.

### Two of your six items were already fixed on the base branch

I took the base's version of those files rather than yours. Nothing of yours was lost in either case.

- **Line endings.** #473 landed a `.gitattributes` with the same `* text=auto eol=lf` rule, plus binary declarations and the reasoning, and normalized the same four CRLF files. Your two font LICENSEs were already byte-identical to it.
- **Advisories.** #425 and #440 cleared xmldom, fast-uri and sharp. The base is on wrangler 4.131.1, above the `^4.131.0` here, #424 is closed, and `npm audit` against the base lockfile reports 0 vulnerabilities.

The diff is 10 files instead of 17, and `Closes #415` and `Closes #424` are gone from the body.

### Your three remaining fixes are intact

I removed each one and watched the right test fail:

- **#416** - reverting to the old `__PATH__executable__` literal fails 4 of 4 tests. Dropping either platform arm fails only that arm's test.
- **#423** - removing your `hitArea` line makes the new spec fail at the second click, while its neighbour still passes. It is a real regression test, and the geometry holds up: both `drawGraphic` functions keep the checked shape inside the unchecked one, so the bounds cached in the constructor stay valid.
- **#419** - reverting the `data/output` clause fails all four dataset cases and leaves all four controls green.

### Three things I changed

CodeRabbit raised the first two.

1. `bash scripts/ci-changes.test.sh` ran on the line above `vp test` in a single step. GitHub runs a step with `bash -e`, so a failure there exited before `vp test` and the run reported nothing at all about the 569 unit tests. It is now `scripts/ci-changes.test.mjs`, which `vp test` finds through the `scripts/**/*.test.mjs` glob the config already had for `localpreview.test.mjs`. Same cases, plus a guard so an extraction that matches nothing fails loudly instead of passing everything against empty input. It also runs for anyone running `vp test` locally now.

2. `factorioReadData` resolves symlinks before searching. If you read the review thread, note that CodeRabbit's suggested diff imports `realpathSync` from `node:path`, which has no such export. It is in `node:fs`.

3. My rewrite then went red in CI, and yours would not have. Linux caps a single argv or envp string at 128 KiB, and I was passing the 232 KB file list through the environment. macOS has no such cap, so it was green on my machine and every smaller case passed in CI. It goes on stdin now. Your shell version never met this because it kept `files` in one process and never spawned anything.

### Where it stands

Green on `289a0e02`: checks, Rust and all four Playwright shards.

Does this look right to you? I am happy to put any of it back if I have misread something, particularly the two dropped halves in case there was something in them I did not spot.

PR #431, opened 2026-09-11
Fix ALT toggle label contrast in both states

### LINKED ISSUE #429: [BUG] The ALT button's label is unreadable while the button is on (1.06:1 contrast)
**Describe the bug**

The ALT button in the ToolsPanel turns amber when entity info is on. Its label keeps the same light grey it uses when the button is off. Light grey on amber is almost the same brightness, so the word "ALT" nearly vanishes. The off state reads fine; the on state does not.

**Measured**

Colours read from the source, then confirmed against real rendered pixels in a headless Chromium run at 1280x720 (I sampled the 36x36 ALT cell out of a screenshot; both states have the same 17 pure label pixels).

| State | Label | Fill | Contrast | WCAG AA (4.5:1) |
| --- | --- | --- | --- | --- |
| ALT on (entity info visible) | `#cccccc` | `#f1be64` | **1.06:1** | fail |
| ALT off | `#cccccc` | `#646464` | 3.68:1 | fail |

The label is 14px at weight 300 (`styles.controls.checkbox`, `style.ts:67-72`), which WCAG counts as normal text, so it needs 4.5:1.

Where the three colours come from:

- label - `colors.controls.checkbox.foreground.color` = `0xcccccc`, `packages/editor/src/UI/style.ts:18`
- on fill - `ALT_ACTIVE_COLOR` = `0xf1be64`, `packages/editor/src/UI/ToolsPanel.ts:69`
- off fill - `colors.controls.button.background.color` = `0x646464`, `packages/editor/src/UI/style.ts:13`

`addToggleHighlight` (`ToolsPanel.ts:93-98`) inserts the amber rect at index 1, right above the slot background and below the content. So the label draws over a solid amber fill with nothing blended into it. That placement is deliberate and correct for the hover and press feedback it was written for - the problem is only the label colour on top of it.

**To reproduce**

1. Open the editor at 1280x720. Entity info is on by default, so the ALT button starts amber.
2. Look at the ALT button in the ToolsPanel, bottom right.
3. Press <kbd>Alt</kbd>. The button goes grey and the label becomes readable again.

**Expected behavior**

The label stays readable in both states.

**Fix**

Flip the label to black while the highlight is visible. Measured against the two options:

| Change | ALT on | ALT off |
| --- | --- | --- |
| Invert the label to `#000000` only while the highlight is on | **12.29:1**, passes AAA | 3.68:1, unchanged |
| Always `#000000` | 12.29:1, passes AAA | 3.55:1, slightly worse than today |

So the invert is the better of the two. Always-black fixes the on state and makes the off state a little worse, because black on `#646464` is darker than the grey label it would replace.

**The off state is a second, smaller problem**

3.68:1 also fails AA. It is readable, so it is not what this issue is really about, but it is cheap to fix in the same pass. A `#ffffff` label gives 5.92:1 on the grey. On its own that does not help the amber (1.71:1), so it only works next to the invert above.

**Siblings**

`addToggleHighlight` has exactly one call site, so ALT is the only slot with a persistent colour overlay. But the same light grey label sits over `Button`'s own fills, which use `#b16925`:

- active, alpha 1: 2.66:1
- hover, alpha 0.5 over `#646464` (blends to `#8a6644`): 3.22:1

Those reach `DescribedButton` and `ToggleControl` labels too. They are momentary rather than persistent, so they are far less visible than the ALT case, but the same idea would fix them.

byalex33:
The ALT label now uses black while entity info is enabled and white while disabled, giving 12.29:1 contrast on amber and 5.92:1 on grey. The existing ticker keeps the label and highlight synchronized after either keyboard or mouse toggles; the label clones its style so other controls are unaffected.

Adds a rendered-pixel regression test covering initial state, keyboard toggling, mouse toggling, and toggling back off.

Validation: `vp check --fix` passed; `vp test` passed all 444 tests after locally normalizing the fixture line endings affected by existing issue #415; all five `tests/tools-panel.spec.ts` browser tests passed. Also visually checked both states and mouse/keyboard interaction in the built-in browser at 1280×720.

Closes #429.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the ALT toggle’s text contrast so its label remains readable in both active and inactive states.
  * Updated text tool icons to match the checkbox styling when their fill color changes.

* **Tests**
  * Added coverage verifying ALT label colors across button states and focus interactions.

### DIFF
--- a/packages/editor/src/UI/ToolsPanel.ts
+++ b/packages/editor/src/UI/ToolsPanel.ts
@@ -61,7 +61,8 @@ class ActionSlot extends Slot<undefined> {
  * icons that do exist).
  */
 function createTextIcon(text: string): Container {
-    const label = new Text({ text, style: styles.controls.checkbox })
+    const label = new Text({ text, style: styles.controls.checkbox.clone() })
+    label.style.fill = 0xffffff
     label.anchor.set(0.5)
     return label
 }
@@ -287,6 +288,9 @@ export class ToolsPanel extends Panel {
         const altHighlight = addToggleHighlight(altSlot, ALT_ACTIVE_COLOR)
         this.altHighlightTick = () => {
             altHighlight.visible = G.BPC.overlayContainer.entityInfoVisible
+            if (altSlot.content) {
+                altSlot.content.tint = altHighlight.visible ? 0x000000 : 0xffffff
+            }
         }
         G.app.ticker.add(this.altHighlightTick)
 
--- a/tests/tools-panel.spec.ts
+++ b/tests/tools-panel.spec.ts
@@ -65,6 +65,53 @@ test.beforeEach(async ({ page }) => {
     await waitForEditor(page)
 })
 
+test('ALT label is black on amber and white on grey (#429)', async ({ page }) => {
+    await page.setViewportSize({ width: 1280, height: 720 })
+    const bounds = await page.evaluate(() => window.__fbe_test.toolsPanelBounds())
+    const x = bounds.x + 12
+    const y = bounds.y + 12
+    const expectColors = async (foreground: number[], background: number[]) => {
+        await expect
+            .poll(async () => {
+                const png = await page.screenshot({
+                    clip: { x: x + 4, y: y + 8, width: 28, height: 20 },
+                })
+                return page.evaluate(
+                    async ({ data, foreground, background }) => {
+                        const image = new Image()
+                        image.src = `data:image/png;base64,${data}`
+                        await image.decode()
+                        const canvas = document.createElement('canvas')
+                        canvas.width = image.width
+                        canvas.height = image.height
+                        const context = canvas.getContext('2d')
+                        if (!context) throw new Error('2D canvas unavailable')
+                        context.drawImage(image, 0, 0)
+                        const pixels = context.getImageData(0, 0, canvas.width, canvas.height).data
+                        return [foreground, background].every(color => {
+                            for (let i = 0; i < pixels.length; i += 4) {
+                                if (color.every((channel, j) => pixels[i + j] === channel))
+                                    return true
+                            }
+                            return false
+                        })
+                    },
+                    { data: png.toString('base64'), foreground, background }
+                )
+            })
+            .toBe(true)
+    }
+
+    await expectColors([0, 0, 0], [241, 190, 100])
+    await page.keyboard.press('AltLeft')
+    await expectColors([255, 255, 255], [100, 100, 100])
+    await page.mouse.click(x + 18, y + 18)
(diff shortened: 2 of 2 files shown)


wormeyman:
Verified this independently against a worktree at `58fd9f13`, with the dev server running the PR head rather than the base. Both of your numbers are exact, and the two states you did not mention hold up as well.

Measured contrast in all four states a user can reach:

| State | Text | Background | Ratio |
| --- | --- | --- | --- |
| ALT on, idle | `0x000000` | `0xf1be64` | 12.29:1 (AAA) |
| ALT on, hovered | `0x000000` | `0xd19344` | 7.97:1 (AAA) |
| ALT off, idle | `0xffffff` | `0x646464` | 5.92:1 (AA) |
| ALT off, hovered | `0xffffff` | `0x8a6644` | 5.17:1 (AA) |

The hover rows are worth having because `Button` composites its rollover sprite over the highlight, so the amber under the label while hovered is not `ALT_ACTIVE_COLOR`. Both still clear AA.

I mutation-checked the new test, since a rendered-pixel assertion can pass for the wrong reason:

- Drop the `tint` block and it fails at line 105, the ALT-on black-on-amber check.
- Drop `label.style.fill = 0xffffff` and it fails at line 107, the ALT-off white-on-grey check.

Each half of the fix is pinned by a different assertion, which is what you want. The second one is also the only thing holding the fill change: a black tint over any fill is still black, so the ALT-on state renders byte-identically without it. I measured that rather than assuming it.

I also diffed the whole panel before and after. 111 of 23,072 pixels change, all of them inside one 24x11 box at the ALT glyphs, so the other eight slots come out byte-identical.

Two things I checked that are fine, so nobody re-opens them later: `Container.tint` does exist in pixi 8.19.0, so the `Container`-typed `content` needs no cast; and the per-frame write is free, because the tint setter returns early on an unchanged value (`if (bgr === this.localColor) return`).

One non-blocking suggestion. This spec's header comment explains why `PADDING`/`REPLACE_Y` and `ROW_BUTTON_WIDTH`/`ROW_BUTTON_HEIGHT` are imported instead of copied: a hand-written coordinate desyncs silently, and the test then fails by sampling the wrong pixels instead of by a clear assertion. The new test carries the same class of literal. `bounds.x + 12` copies `slotsContainer.position.set(12, 12)`, and `[241, 190, 100]` and `[100, 100, 100]` copy `ALT_ACTIVE_COLOR` and `colors.controls.button.background.color`. The two colours are already importable from `style.ts`, so that part is cheap; the offsets would need an export. Your call whether it earns the churn.

Nice fix, and thank you for adding the pixel test rather than eyeballing it.

Issue #206, opened 2026-08-11
[FEAT] Better zoom levels


wormeyman:
**Is your feature request related to a problem? Please describe.**
It would be great if we could have zoom levels the same as the game.

**Describe the solution you'd like**
When I scroll my mouse it would zoom in and out the same as the game, being able to zoom a bit further in our out is fine.

**Describe alternatives you've considered**
None

**Additional context**
https://lua-api.factorio.com/latest/concepts/ZoomSpecification.html
https://wiki.factorio.com/Debug_mode

wormeyman:
Measured the editor's side of this and read the game's, so the work has numbers under it. Short version: the units already agree, so this is a step-and-limits problem rather than a conversion problem, and there are three separate defects behind "doesn't feel like the game".

## The units already match, which is the good news

`EntityContainer.position` multiplies world coordinates by 32 (`EntityContainer.ts:296`), so the editor draws 32 px per tile at scale 1. Factorio's runtime API at 2.0.77 says of `LuaPlayer::zoom`:

> The baseline zoom level is 1. Values greater than 1 will zoom in closer to the world and values between 0 and 1 will zoom out away from the world.

Same baseline, same pixels per tile. **The editor's scale number is already Factorio's zoom number**, so nothing has to be converted. Anything below is about which values we step through and where we stop.

## Defect 1: the step is asymmetric, so zoom does not round-trip

`BlueprintContainer.zoom()` uses a flat `zoomFactor = 0.1` and calls `viewport.zoomBy(±0.1)`. `Viewport.zoomBy` does `this.scaleX += delta`, and `scaleX` is reset to 1 after every `_updateMatrix`, so the value multiplied into the transform is `1 + delta`.

That makes zoom in `× 1.1` and zoom out `× 0.9`. Those are not inverses:

```
1.1 × 0.9 = 0.99
```

Scroll one notch in and one notch out and you are 1% further out than you started. Ten round trips lose about 10%, and there is nothing that snaps back. A geometric step needs `× k` and `× 1/k` - with k = 1.1, out should be `× 0.909…`, not `× 0.9`.

## Defect 2: the wheel throws away how far you scrolled

```js
if (Math.sign(e.deltaY) === 1) {
    this.zoom(false)
} else {
    this.zoom(true)
}
```

[...]

PR #442, opened 2026-09-16
Confirm WHEEL_NOTCH_PX against a trackpad, unchanged

### LINKED ISSUE #216: Tune WHEEL_NOTCH_PX against real scrolling hardware
`WHEEL_NOTCH_PX = 100` in `packages/editor/src/core/zoomLevels.ts` is the **only unmeasured number** in the zoom work from #206. Everything else in that module came off the game: the `2^(1/7)` step, the off-rung rule, both limits.

Since the wheel went continuous it no longer controls chunkiness. It sets exactly one thing: **how far a given amount of scrolling travels**.

## The numbers as they stand

- one rung per 100px of wheel travel, and one rung is +10.4%
- the full range is `log2(3/0.1) x 7` = **34.35 rungs**, so crossing it end to end costs **3435px** of scrolling - about 34 notches on a notched mouse
- a notched mouse in Chrome sends `deltaY` 100 per notch, so one notch is exactly one rung and that hardware walks the game's ladder

## Why it cannot be settled by reasoning

The game's cadence is known - measured in `tools/oracle/fixtures/zoom-limits.json`, it moved **11 rungs in 11 ticks**, the whole top of its range in 183ms. But that is per *time*, and converting it to pixels needs to know how many pixels a given device emits per second during a flick. A Mac trackpad and a notched mouse differ by an order of magnitude here.

## What would settle it

One line from the console probe, run at localhost:8080 after the editor loads. It groups wheel events into gestures and prints events, pixels and rungs per gesture:

```js
(() => {
    let events = 0, px = 0, startZoom = null, idle = null
    const zoom = () => window.__fbe_test.viewportScale()
    const rung = v => Math.log2(v) * 7
    addEventListener('wheel', e => {
        if (startZoom === null) startZoom = zoom()
        events++; px += e.deltaY
        clearTimeout(idle)
        idle = setTimeout(() => {
            const end = zoom()
            console.log(`${events} events  ${Math.round(px)}px  ->  ${(rung(end) - rung(startZoom)).toFixed(2)} rungs   zoom ${startZoom.toFixed(3)} -> ${end.toFixed(3)}`)
            events = 0; px = 0; startZoom = null
        }, 200)
    }, { capture: true })
    console.log('armed - scroll and watch')
})()
```

Do one gentle scroll and one fast flick, with whatever hardware is normally used, and paste both lines.

## Cost

One line. The unit tests reference the symbol rather than the literal, and `tests/zoom-ladder.spec.ts` imports it, so both follow automatically.

Note this is feel, and no passing test can see it - the same shape as the rail signal snap radius, which shipped fully mutation-checked at a value that left the signal 390px from the pointer (#143).

wormeyman:
## Summary

A human-in-the-loop session with a notched mouse (MX Master, ratchet
mode) measured the one number in `zoomLevels.ts` that #206 left
unmeasured, via the console probe from #216:

| Gesture | Events | Total px | Rungs moved | Zoom |
|---|---|---|---|---|
| Light scroll | 164 | 54px | -0.54 | 0.533 → 0.505 |
| Fast flick | 81 | 2,145px | -20.20 (hit the floor) | 0.739 → 0.100 |

At the current `WHEEL_NOTCH_PX = 100`, the light scroll lands under one
rung - controlled. The flick carries enough pixels to reach `ZOOM_MIN`
in one gesture. Both felt right, so this is a no-op on the constant -
only the comment changes, from "never measured" to what was measured.

**A secondary finding, flagged but not solved here:** the file's first
paragraph claims a notched mouse sends a flat 100px per notch in
Chrome. This session's actual notched-mouse capture didn't look like
that - it was a continuous stream from 1px up past 100px within a
single flick, the same shape a trackpad produces. Whether that
100px/notch premise holds on macOS at all is untested and worth its
own look; the comment says so rather than asserting either way.

(An earlier round of this PR mislabeled the hardware as a trackpad -
it was this same mouse, tested before the ratchet/notched distinction
was clear. Corrected.)

## Test plan

- [x] `vp check packages/editor/src/core/zoomLevels.ts` passes
- [x] No test asserts the literal; all reference the `WHEEL_NOTCH_PX` symbol, so none needed updating

Closes #216

https://claude.ai/code/session_013pLo2kkHe3xReZyPV2NJhM

## Summary by CodeRabbit

* **Documentation**
  * Updated wheel-zoom documentation with measurements from driving-session testing.
  * Added examples showing how scroll distances translate to zoom changes, including large trackpad flicks reaching the minimum zoom level.
  * Clarified zoom-range clamping behavior for exceptionally large trackpad inputs.
  * Documented that the shipped wheel-zoom notch setting remains unchanged.
  * macOS-specific behavior remains untested.


wormeyman:
Paused, and the commit/PR body need correcting before this can merge.

Two problems, found while redoing the session:

1. **Wrong hardware.** The first round was scrolled with a mouse (MX Master, notched/ratchet mode), not a trackpad. The commit and this PR currently say "trackpad" throughout - that provenance is wrong.

2. **The redo contradicts the file's other claim about notched mice.** `zoomLevels.ts` opens this comment with "`WheelEvent.deltaY` is 100 per notch for a notched mouse in Chrome" - that's stated as fact, not measured in this repo. Redoing the probe with per-event `deltaY` logging, on the actual notched MX Master, on macOS:

   - Gentle: 140 events, 189px total, deltas like `[1,1,1,1,2,1,1,1,1,1,1,1,2,2,...]`
   - Flick: 81 events, 2145px total, deltas ramping `[1,1,2,...,91,114,101,106,104,...]`
   - Three more gestures in the same session, all similarly continuous rather than chunky

   None of that looks like discrete ~100px clicks. It looks like macOS is smoothing/accelerating the notched wheel's input before Chrome ever sees it - the same shape a trackpad produces, just a different acceleration curve. If that holds up, the "100px = 1 notch" premise the whole file's design rests on may not be true on macOS at all, which is a bigger question than the original issue (tuning `WHEEL_NOTCH_PX`) and needs its own look before this PR's conclusion can be trusted.

Leaving this in draft. Next session: settle whether "notched mouse = 100px/notch in Chrome" holds on macOS before re-drawing any conclusion about `WHEEL_NOTCH_PX` itself.

PR #450, opened 2026-09-18
Canonicalise the proxy target hostname so a trailing dot cannot pass a denial

byalex33:
## Summary

`checkProxyTarget` compared hostnames as strings, and the WHATWG parser keeps a trailing root label on a name. So `localhost.`, `printer.local.` and `metadata.google.internal.` were all allowed by `/corsproxy` where the undotted form is refused 403.

IPv4-shaped hosts are slightly different: the parser removes exactly one trailing empty label, so `127.0.0.1.` was already refused - but `127.0.0.1..` and `169.254.169.254..` survive as-is, fail the literal regexp, and reached the catch-all.

## Fix

`canonicalHostname()` lowercases and strips every trailing dot, once, at the top of `checkProxyTarget`. Every rule runs against that one spelling, including the `selfHostname` side of the anti-loop check (the arrival hostname is whatever the client put in `Host`). The URL that gets fetched is rewritten to the same spelling so the guard and the request cannot disagree on what the target is.

A public name with a trailing dot is still accepted - it is a legal spelling - but it is fetched without the dot.

## Tests

Ten new cases in `tests/corsproxy.test.ts`: the dotted spellings of every named denial, the two-dot IPv4 forms, `https://./x`, the dot on the arrival-hostname side, and two acceptance cases (public host and allowlisted host with a trailing dot, asserting the fetched `href` is undotted).

Measured: all ten fail against the previous `proxyTarget.ts` and pass with this one. `vp check` clean, 37/37 in the file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hostnames with trailing dots are now normalized consistently before proxy requests.
  * Prevented trailing-dot hostname variants from bypassing hostname restrictions.
  * Continued blocking unsafe targets, including localhost, private addresses, metadata endpoints, and invalid hostnames.
  * Improved handling of trailing dots in deployment host headers and allowlisted hosts.


wormeyman (commented):
@byalex33 I ran the real module at both commits instead of reading the diff, and imported `checkProxyTarget` from `718dbcc7` and from `ec0f4477` over the audit's own case list.

Eleven of twenty-two cases change, and every one moves the right way:

| target | `718dbcc7` | this PR |
|---|---|---|
| `localhost.` | allowed | deny 403 |
| `printer.local.` , `foo.internal.` | allowed | deny 403 |
| `metadata.google.internal.` | allowed | deny 403 |
| `127.0.0.1../x` , `169.254.169.254../x` | allowed | deny 403 |
| `https://./x` , `https://../x` | allowed | deny 403 |
| this deployment, dotted | allowed | deny 403 |
| the dot on the arrival side | allowed | deny 403 |
| every undotted control | deny 403 | deny 403, unchanged |
| ordinary public host | allowed | allowed, fetched undotted |

The ten new cases fail against the old module, all ten and no others, and 37/37 pass with the new one. Your description matches what I measured.

Two things came out of it that the description does not mention. Neither is a defect in what you wrote.

**1. One case ends up allowed where it used to be refused.** `https://pastebin.com.:8443/raw/abc`:

```
718dbcc7: deny 403  Only the default https port is proxied for unrecognised hosts
ec0f4477: ALLOW (allowlisted)  https://pastebin.com:8443/raw/abc
```

That comes from ordering rather than from anything in `canonicalHostname`. Stripping the dot makes the name match `ALLOWED_HOSTS`, and the allowlist arm returns before the port check runs. So the dotted spelling now behaves the same as the undotted one, which was already allowed on 8443 for the same reason. Consistency is presumably the intent, but it is a real change in what the route accepts and it is not in the description. If the allowlist should still respect the port rule, that is its own change.

**2. It does not close the self-proxy gap, and the two are easy to confuse.** Measured at this head, `https://fbeworkeyman.wormeyman.workers.dev/corsproxy?url=x` is still an accepted target. Only its dotted spelling was normalised. The guard compares one arrival hostname, so every other hostname this deployment serves still passes.

Worth doing in the same file at some point, either by refusing `CUSTOM_ORIGIN`, `LEGACY_HOSTNAME` and a `.workers.dev` suffix, or by setting a marker header on the outbound fetch and refusing any inbound request that carries it, which also covers hostnames nobody remembered to list. I mention it because after this merges, "canonicalise the hostname" looks finished and that one can disappear with it.

PR #451, opened 2026-09-18
Give every Worker-built Response a header policy, and stop /corsproxy re-emitting the upstream content-type

byalex33:
## Summary

Two findings on the same route, fixed together because they are one mechanism.

**`/corsproxy` re-emitted an attacker-chosen `Content-Type` over an attacker-chosen body.** The catch-all in `proxyTarget.ts` accepts any public https host, and `content-type` was on the passthrough list, so a visitor who navigated to a crafted `/corsproxy?url=` link got the upstream's bytes rendered as whatever media type it declared, under `fbe.factorygamefan.com`. (`Headers.set` refuses CR/LF/NUL, so the header could not be split - the value was the whole problem.)

**No Response the Worker builds itself carried any header policy.** `public/_headers` is read only by the asset router; `/corsproxy` is on `run_worker_first`, so its Response - and every text error and the legacy-hostname redirect - went out exactly as constructed. The one route serving third-party bytes was the one route with no `nosniff`, no `Referrer-Policy` and no CSP.

## Fix

New pure module `packages/worker/src/responseHeaders.ts`, arranged like `proxyTarget.ts` and for the same reason (nothing in `packages/worker` is collected by `vp test`; a module free of Cloudflare types can be imported from `tests/`):

- `ownHeaders()` - the policy on every Worker-built Response: `nosniff`, `strict-origin`, and a CSP stricter than the app's own (`default-src 'none'; frame-ancestors 'none'; sandbox`) - nothing the Worker builds is a document anyone should run.
- `proxyResponseHeaders()` - the proxy's header rebuild. `content-type` is off the passthrough list; the response carries a fixed `text/plain; charset=utf-8` and `Content-Disposition: attachment`.

Behaviour-neutral for the editor: every consumer in `bpString.ts` reads with `r.text()` / `r.json()`, neither of which consults `Content-Type`, and `fetch()` ignores `Content-Disposition`.

Also: the redirect is built by hand (`Response.redirect`'s headers are immutable), and the 413 path cancels the upstream body it declines to relay.

## Verification

Measured under `wrangler dev --local`:

| Route | Before | After |
|---|---|---|
| `/corsproxy?url=https://example.com/` | `text/html`, no policy headers | `text/plain; charset=utf-8`, `nosniff`, `strict-origin`, CSP, `attachment` |
| `/corsproxy` (400) | `text/plain` only | + all three policy headers |
| `/` | `_headers` CSP | unchanged - asset route untouched |

`tests/worker-response-headers.test.ts`: 11 tests on the pure module plus a source scan asserting `index.ts` never uses `Response.redirect` and routes every `new Response(` it sends through the module. The scan fails on both counts against the previous `index.ts`. `vp check` clean; the Worker type-checks under its own tsconfig.

Related: #450 (hostname canonicalisation, same file family).

## Summary by CodeRabbit

* **Security**
  * Improved security protections for generated and proxied responses, including stricter content handling, referrer policies, and content security policies.
  * Responses now prevent caching and limit exposed upstream headers.

* **Bug Fixes**
  * Oversized upstream responses are safely terminated before returning an error.
  * Legacy-host redirects now use consistent, controlled response headers.
  * Proxied responses consistently use plain-text downloads and deployment-origin access controls.


wormeyman (commented):
@byalex33 I drove the real `fetch` handler with a stubbed global `fetch`, so the whole `/corsproxy` path runs (guard, size cap, header rebuild) with no network, at `718dbcc7` and at `dd39717f`.

| upstream sends | `718dbcc7` | this PR |
|---|---|---|
| `text/html` plus `set-cookie` | `text/html`, four headers, no policy | `text/plain; charset=utf-8`, `attachment`, `nosniff`, `strict-origin`, CSP |
| nothing at all | no `content-type` at all | `text/plain; charset=utf-8` plus `nosniff` |
| a 403 refusal | `content-type` only | full policy |
| the legacy-hostname 301 | `location` only | full policy |

The second row is the one I would keep. At base, an upstream that declares nothing leaves the proxy emitting nothing, which is the shape a browser sniffs.

I checked the behaviour-neutrality claim instead of taking it. `x-ratelimit-remaining` at `bpString.ts:321` is the only response header anything in `packages/editor` or `packages/website` reads, and it stays on the passthrough set. Every body consumer uses `.text()` or `.json()`. Nothing navigates to `/corsproxy`. Full unit suite 455/455 with this applied.

The new test file holds up under mutation. Three tried, three caught: dropping the fixed `content-type` set (2 failures), putting `content-type` back on the passthrough list (1), and making `textResponse` stop using `ownHeaders` (1, through the source scan). Worth noting for whoever reads it next that the second of those is behaviourally inert, since the explicit `set` runs after the passthrough loop. That is exactly why pinning the set rather than the output is the right test.

One suggestion, take it or leave it. The zone has no response-header transform rule and no page rules at all (we read it today), so this Worker is the only thing that can put a header on that route, which makes it the only place an `X-Robots-Tag` can go. Over a 24 hour window Googlebot accounted for 2,940 `/corsproxy` requests from seven addresses, and 3,590 counting the other Google ranges, out of 40,470 total. That is 8.9% of the route. There is no `robots.txt` in this repository at all. One more entry in `OWN_RESPONSE_HEADERS` plus a `packages/website/public/robots.txt` carrying `Disallow: /corsproxy` would remove most of it, and this PR is already the change that decides what those responses carry.

No concerns otherwise.

wormeyman:
@byalex33, a note for the morning: this and #450 are merged, and we picked up the follow-ups tonight so they did not have to wait.

A correction to what we suggested here. We asked for `X-Robots-Tag` in `OWN_RESPONSE_HEADERS` plus a `robots.txt`, and only the `robots.txt` half does anything. A second read of the logs showed that no page links to `/corsproxy`. The editor builds that URL in script when a `?source=` link loads, and `/` and `/corsproxy` ran almost exactly 1:1, so Google's renderer is most likely running the editor and fetching through the proxy. `X-Robots-Tag` only governs indexing and would not stop those fetches; `Disallow: /corsproxy` does. That is #453, so nothing needs adding here.

Separately, `wrangler.jsonc` never set `preview_urls`, so every deployed version (265 of them) stayed live on its own `<version>-fbeworkeyman.wormeyman.workers.dev` hostname, running its own code. Your fixes in this PR and #450 could not reach those hostnames, and the versions from before #272 still run the proxy with no target guard and `Access-Control-Allow-Origin: *`. #452 turns previews off.

Also opened tonight, if you want to look: #454 (CLAUDE.md Node pin, `.gitignore` for `.dev.vars`), #455 (`?source=` matched by exact key), #459 (null-prototype `FD` collections plus a rollback when `initBP` throws), and issue #456, where the self-proxy guard only refuses the arrival hostname. #456 follows on naturally from #450 if you want to take it.

PR #474, opened 2026-09-20
Make the exporter work on Linux, and download the Space Age build

wormeyman:
Three changes in `packages/exporter`, and one job removed from CI.

## The two `#[cfg(target_os)]` blocks become a runtime match

`download()` had one block for the Windows zip and one for the Linux tar.xz. Those two were the exporter's entire compile-time platform surface; everything else already matches on `std::env::consts::OS` at run time.

Gating the call sites bought nothing. `zip`, `tokio-tar` and `async-compression` are plain unconditional entries in `Cargo.toml`, so Linux was already compiling the zip crate and Windows the tar crates. Only the call sites were hidden.

Two things follow:

- Every target type-checks both arms, which is what `rust-windows` existed to provide. The job is deleted, and every remaining job runs on `ubuntu-latest`.
- `stream` and `out_dir` stop being unused on macOS, so `cargo clippy --locked -- -D warnings` goes from 2 errors to 0 there, matching Linux.

## `zip` keeps only deflate

Read from the published archive's own central directory over two HTTP range requests rather than guessed: all 8,069 entries in `factorio-demo_win_2.0.68.zip` use deflate and nothing else.

`default-features = false, features = ["deflate"]` therefore drops 30 crates from `Cargo.lock`, among them `zstd`, `zstd-sys`, `aes`, `bzip2`, `sha2`, `pbkdf2` and `ppmd-rust`.

## The download asks for `expansion`, and finds its own version

The URL hardcoded `alpha`, which is the base game. That one word is why this path could not produce Space Age data. It now asks for `expansion`, and the account behind `FACTORIO_TOKEN` has to own Space Age.

The version came from `static FACTORIO_VERSION: &str = "2.0.68"`, which nothing updated. `resolve_version()` now reads `stable.expansion` from <https://factorio.com/api/latest-releases>, and `FACTORIO_VERSION` still pins a specific one. The API reports 2.0.77 today, which is the version the oracle fixtures were recorded at.

## A Linux basisu, so sprite compression works there too

Sprite compression shelled out to a bare `./basisu`, which is the macOS arm64 build; on Linux that is simply the wrong binary. `basisu_for()` now picks per platform, and Windows is named explicitly rather than relying on CreateProcess appending the extension.

The Linux binary is upstream's own x86-64 build from the 1.16.4 release. That is the same archive whose `basisu.exe` is byte-identical to the one already tracked here, so it is the provenance the repo already trusts.

**The three are not interchangeable, and the committed textures came from the macOS build.** The same PNG gives 139,208 bytes on macOS and 139,094 on Linux, and 7 of 8 sprites differed. The pictures match: alpha is bit-identical across the sheet, transparent pixels are identical, and the mean channel difference is 0.226 of 255. The bytes do not match, so regenerating on Linux would rewrite all 3,060 `.basis` files, about 152 MB, with nothing visible in it.

The committed output is therefore left alone, and both CLAUDE.md and `.github/workflows/README.md` say so rather than leaving it to be discovered. If that rewrite is ever wanted, the Linux output is reproducible: upstream's x86-64 build gives the same sha256 on native x86-64 hardware and under emulation on an arm64 Mac, measured both ways.

A test pins the mapping and checks all three binaries are tracked, because a dropped one fails only on the platform nobody happens to be testing on.

## What this stops covering

A Windows cross-check is still unavailable, and the reason moved rather than went away. Dropping zip's defaults removed `zstd-sys`, the crate `.github/workflows/README.md` used to name, and `liblzma-sys` took its place, because `async-compression` needs it for the Linux `.tar.xz`. That really is the format: the URL serves `factorio-demo_linux_2.0.68.tar.xz`. Measured from macOS, `cargo check --target x86_64-pc-windows-msvc` stops at `'stdlib.h' file not found`, since no Mac or Linux host carries the Windows SDK headers.

So nothing checks that the exporter's dependency tree still builds on windows-msvc any more. A crate bump could break `npm run start:exporter` for a Windows contributor while CI stays green. Both READMEs say so rather than leaving it implied.

## Checks

`vp check` clean. On macOS, `cargo build --locked`, `cargo test --locked` (9 passed), `cargo fmt --check` and `cargo clippy --locked -- -D warnings` are all clean, with 10 tests passing.

One limit worth naming: the download path is not exercised here, because it needs credentials this machine does not have. The `expansion` and version changes were verified read-only, against the public API and the published archives.

@claude review

https://claude.ai/code/session_01SVXZymns1HCNN2GRrWQwrF


wormeyman:
@claude Review

PR #475, opened 2026-09-20
Stop caching 1.2 GB of debug info the Rust job never reads

wormeyman:
The Rust job was spending most of its time moving a cache, and almost none of it compiling.

Measured on run [35543940158](https://github.com/FactoryGameFan/factorio-blueprint-editor/actions/runs/35543940158), 2026-09-20:

| step | time |
| --- | --- |
| Cache cargo (restore) | 49s |
| Post Cache cargo (save) | 19s |
| Build + Test + Format + Clippy | 13s |
| everything else | 9s |
| **total** | **90s** |

The cache itself was 1,884 MB.

## Where the size came from

Debug info. Measured locally the same day: a clean build directory for this crate is 1.7 GB, with 1.2 GB of that in `target/debug/deps`. Building with `CARGO_PROFILE_DEV_DEBUG=false` and `CARGO_INCREMENTAL=0` gives **506 MB** instead, a 70% cut, and the build does not get slower - 15.6s against 20.3s on the same machine. Cargo confirms it applied by printing `[unoptimized]` where it used to print `[unoptimized + debuginfo]`.

Nothing in this job opens a debugger, so that information was being compressed, uploaded, downloaded and decompressed for nobody.

Both are set as job-level `env:` rather than in `Cargo.toml`, so anyone building locally keeps backtraces with line numbers.

## The cache key has to move with it

`actions/cache` does not re-save on a hit. Leaving the key alone would have restored the old 1.9 GB cache for its lockfile forever and never replaced it, so the change would have looked like it did nothing. Hence `cargo-v2-`, and any future change to how these artifacts are produced needs the same bump.

## What is deliberately not here

The Windows job keeps its old key and settings. PR #474 deletes that job, and editing a job another branch removes is a merge conflict for no gain.

This also started as a look at whether CI should move into a prebuilt container image. The numbers said no, and that is worth writing down: a Playwright shard is 390s of which 365s is running specs, Playwright browsers already restore from cache in 0s, and the setup an image could replace is about 14s per job. Pulling a ~2.8 GB image across six jobs would cost more than it saved.

## Checks

`vp check` clean. The workflow parses, and the `rust` job's `env` and cache keys read back as intended.

It is measured on a runner, not estimated. Run [35545415494](https://github.com/FactoryGameFan/factorio-blueprint-editor/actions/runs/35545415494), both attempts:

| | before | after |
| --- | --- | --- |
| cache size | 1,884 MB | 251 MB |
| restore | 49s | 6s |
| save (post step) | 19s | 0s on a hit, 4s when it writes |
| whole job, warm | 90s | **30s** |

The first attempt had no `cargo-v2` cache to restore and paid a full build: 100s, with 68s in `Build` and 19s in `Clippy`. That is what any future key bump will look like, and `README.md` records it so it does not read as a regression later.

@claude review

https://claude.ai/code/session_01SVXZymns1HCNN2GRrWQwrF

## Summary by CodeRabbit

- **CI Improvements**
  - Reduced Rust CI cache size and improved warm-build performance.
  - Updated cache versioning to ensure builds use the new configuration.
  - Disabled development debug information and incremental compilation in CI to reduce cache overhead.

- **Documentation**
  - Added timing measurements and guidance explaining cache behavior, initial rebuild costs, and performance improvements.


wormeyman:
@claude Review

